PR #1, opened 2022-01-03
added project files

HenryQuillin:


j-mateo (approved):
Looks good!
Have you considered adding eslint? Aside from linting, it integrates with the IDEs pretty nicely to auto-format the code according to best practices

HenryQuillin:
> 

Thanks for taking a look! You are right about the linting. It is in the package.json but I think will add a .eslintrc.

AndrewWestberg (approved):

PR #4, opened 2022-01-08
Add npm script to format files using eslint config

newmskywalker:
Adding an npm script so that we always format files the same way :)

Running `npm run format` should format all the files for the project, but we should try to setup our IDE's so that files are always formatted as soon as they're saved﻿

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,8 +21,9 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
+    "test:ci": "npm run test -- --coverage --watchAll=false",
     "eject": "react-scripts eject",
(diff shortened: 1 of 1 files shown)


j-mateo:
This extension automates file formatting pretty nicely:

https://marketplace.visualstudio.com/items?itemName=rvest.vs-code-prettier-eslint

j-mateo:
@HenryQuillin If you have any styling preference, we can adjust that in .eslintrc and reformat before merging this PR

HenryQuillin:
@j-mateo I really don't have any specific styling preferences as long as everything is consistent. I was having trouble with prettier, but using installing and using prettier-eslint seems to work nicely. I resolved some conflicts but seems that more popped up. Should be minor things.

PR #19, opened 2022-01-30
Finished image upload component and implemented in song upload form

HenryQuillin:


package.json
```diff
@@ -7,6 +7,7 @@
     "@emotion/styled": "^11.6.0",
     "@mui/icons-material": "^5.2.5",
     "@mui/material": "^5.2.7",
+    "cloudinary-react": "^1.7.1",
```
j-mateo, line 10: Are we using Cloudinary for the platform in prod? 
Just making sure we've vetted their server before committing to using it

cc: @AndrewWestberg

package.json
```diff
@@ -24,19 +25,12 @@
     "test": "react-scripts test",
     "test:ci": "npm run test -- --coverage --watchAll=false",
     "eject": "react-scripts eject",
+    "lint:fix": "eslint --fix --ext .js,.jsx,.tsx,.ts .",
```
j-mateo, line 28: I wonder if this is doing the same thing as `format`. Did you see different results?

src/components/ImageUpload.tsx
```diff
@@ -1,65 +1,98 @@
...
+//     imageFile: {
+//       name: string;
+//       size: null;
+//       type: string;
+//     };
+//     releaseDate: string;
+//     title: string;
+//     yourRole: string;
+//   }>;
+// }
```
j-mateo, line 30: do we still need these lines?

src/components/ImageUpload.tsx
```diff
@@ -1,65 +1,98 @@
...
-        >
-          { " " }
-          Drag & Drop{ " " }
-        </StyledPaperInput>
-
-        <Typography variant="body1" align="center" color="primary">
-          OR
-        </Typography>
-        { isDragActive ? <p>Drop the files here ...</p> : <p>Drag n drop some files here, or click to select files</p> }
+        { !uploadedFilePublicId ? (
```
j-mateo, line 63: nit: ternary is a bit hard to read. I think a simple `if` with early return might make it more clear

src/components/ImageUpload.tsx
```diff
@@ -1,65 +1,98 @@
...
+              </Box>
+            ) : (
+              ""
+            ) }
+          </div>
+        ) : (
+          <Box sx={ { alignItems: "center", display: "flex", justifyContent: "center" } }>
+            <Image
+              cloudName={ PUBLIC_CLOUDINARY_CLOUD_NAME }
+              publicId={ uploadedFilePublicId + ".png" }
```
j-mateo, line 89: will they always be `png`s?

src/components/ImageUpload.tsx
```diff
@@ -1,65 +1,98 @@
...
+  const [uploadedFilePublicId, setUploadedFilePublicId] = useState<string>();
+  const onDrop = useCallback(acceptedFiles => {
+    const url = `https://api.cloudinary.com/v1_1/${PUBLIC_CLOUDINARY_CLOUD_NAME}/upload`;
+
+    acceptedFiles.forEach(async (acceptedFile: File) => {
+      const formData = new FormData();
+      formData.append("file", acceptedFile);
+      formData.append("upload_preset", PUBLIC_CLOUDINARY_UPLOAD_PRESET);
+
+      const response = await fetch(url, {
```
j-mateo, line 45: Just so I understand it better, is this an unauthenticated endpoint to upload images?
So anyone with the URL can attempt to upload files? If so, are we able to at least restrict it to a domain whitelist?

j-mateo (commented):
Looking good! I really like the reusable image upload component. I would like to see some basic tests for the SongUploadForm that cover the image upload functionality. Let me know if you want to pair on that 🙏🏽

package.json
HenryQuillin: I was trying to find a way to resolve all fixable problems with one command instead of having to fix each lint warning in each file (for example, removing unused imports). This actually did not work as I was hoping it would and I forgot to remove it. I will delete it.

src/components/ImageUpload.tsx
HenryQuillin: I will make a new branch where I will try to set up the Formik validation for the image component. This interface will be useful for that.

src/components/ImageUpload.tsx
HenryQuillin: Agreed. I will change.

package.json
j-mateo: That was the intended purpose of the format script as well. Did that one not fix some of them?

src/components/ImageUpload.tsx
HenryQuillin: Yes, unless you think otherwise. The image had a white background when I uploaded a jpeg, but converting it to a png allowed it to have a transparent background.

package.json
HenryQuillin: it looks like it does everything except removing unused imports.

HenryQuillin: Which is not too big a deal

src/components/ImageUpload.tsx
HenryQuillin: This is an unsigned upload endpoint. I can implement a signed endpoint that should restrict who can upload files.

src/components/ImageUpload.tsx
```diff
@@ -1,66 +1,130 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  setFieldValue?: (field: string, value: any, shouldValidate?: boolean | undefined) => void;
+  setTouched?: (fields: { [field: string]: boolean }, shouldValidate?: boolean) => void;
 }
 
-export const ImageUpload = (props: ImageUploadProps) => {
-  const { setFieldValue } = props;
+export const ImageUpload = ({
+  fileType = "png",
+  dropzoneLabel = "Drag & Drop (Square Image Only)",
```
j-mateo, line 22: Out of curiosity, why square images only?

src/components/ImageUpload.tsx
```diff
@@ -1,66 +1,130 @@
...
+        formData.append("file", acceptedFile);
+        formData.append("upload_preset", String(PUBLIC_CLOUDINARY_UPLOAD_PRESET));
+
+        const response = await fetch(url, {
+          body: formData,
+          method: "post",
+        });
+
+        const data = await response.json();
+        setUploadedFilePublicId(data.public_id);
```
j-mateo, line 44: not for this pr, but we should add an error handler here.  If the call to the cloudinary api fails, we'd probably want to let the user know so they can retry

src/components/ImageUpload.tsx
```diff
@@ -1,66 +1,130 @@
...
-    },
+    multiple: false,
+    onDrop,
   });
-  return (
-    <Box>
-      { }
+  let render = <div />;
+  if (!uploadedFilePublicId) {
+    render = (
```
j-mateo, line 59: Wondering if the render variable is needed.
I think we should be able to just return early and drop the else

```kotlin
if(!uploadedFilePublicId) {
    return (
        ....
    )
}

//if we got here, uploadedFilePublicId is true
return (...)
```

src/components/ImageUpload.tsx
```diff
@@ -1,66 +1,130 @@
...
+            ""
+          ) }
 
-        <Typography variant="body1" align="center" color="primary">
-          OR
-        </Typography>
-        { isDragActive ? <p>Drop the files here ...</p> : <p>Drag n drop some files here, or click to select files</p> }
+          <Image
+            data-testid="uploadedImage"
+            style={ imgLoaded ? { display: "" } : { display: "none" } }
```
j-mateo, line 113: I think that instead of changing the css display value, you could combine this with the logic from the loading indicator.
if imgLoaded is true, show the image, otherwise show the loading icon

j-mateo (approved):
lgtm!
Just some minor comments

j-mateo:
Great job with the tests. I like that they test what the user sees, not our implementation details

src/components/ImageUpload.tsx
HenryQuillin: If you submit a non-square image it gets stretched. Another option could be to crop the image to a square, but that could cut out parts that the user may want.  I thought the best option would be to just ensure the user submits a square image. Album covers are square already, so is this copy even necessary?

src/components/ImageUpload.tsx
HenryQuillin: Created a click up task for this

src/components/ImageUpload.tsx
HenryQuillin: Changed

src/components/ImageUpload.tsx
HenryQuillin: The idea here is that it is waiting until the image is fully loaded and rendered to show it. It needs to start rendering the image in the DOM once it gets a response from Cloudinary. The loading sign replaces the image loading into the DOM, which is why I used CSS.

j-mateo: Ahh that makes sense!

PR #27, opened 2022-02-21
Refactors the project structure to scale more effectively

scandycuz:
Adds Redux and structures the `src/` directory to scale more effectively.

The structure outline is as follows: 

```
- api/
   // Clients used to interact with the API.
- assets/
   // All static resources (images, svgs, etc...).
- common/
   // Utils and type definitions that don't belong in a particular module.
- components/
   // Top level components that are used anywhere in the app and aren't unique to a single page.
- modules/
   // State data grouped into categories defined by how the are used in the app. 
   // Example sub-directory:
   // - song/
   //    - selectors.ts
   //    - slice.ts
   //    - types.ts
- pages/
   // Individual pages in the app. Components that are only used on a singe page should be placed here.
   // Example page:
   // - home/
   //    - Home.ts
   //    - songs/
   //       - Songs.ts
   //       - Song.ts
- theme/
   // Theme customization files.
```

The structure is meant to be fluid, and should change if some things could be organized more effectively.


.eslintrc.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
     "plugin:testing-library/react",
     "plugin:react/jsx-runtime",
   ],
+  ignorePatterns: ["src/assets/*"],
```
j-mateo, line 19: 👌🏽

j-mateo (approved):
💪🏽

PR #29, opened 2022-02-24
Playlist grid view

scandycuz:
- Adds a grid view for playlists


j-mateo:
Task linked: [CU-2f70x91 Playlist grid view](https://app.clickup.com/t/2f70x91)

scandycuz:
Marking as a draft until back-end playlist data structure is confirmed.

src/pages/home/Home.tsx
```diff
@@ -54,17 +55,19 @@ const Home: FunctionComponent<HomePropTypes> = (props) => {
...
         data-testid="contentBox"
         justifyContent="center"
-        maxHeight={ height && height - 295 }
+        maxHeight={ height && height - 290 }
         overflow={ "scroll" }
         id="content"
       >
         <TabPanel value={ value } page={ Page.songs }>
-          <Box sx={ { marginLeft: "auto", marginRight: "auto", width: "1060px" } }>
+          <Container maxWidth="lg">
```
HenryQuillin, line 63: I am curious to know what the difference is between Container and Box

src/components/SquareGridCard.tsx
```diff
@@ -0,0 +1,40 @@
+/**
+ * A card used with the MUI Grid component that retains a square shape.
+ */
+
+import { Box, Card, CardProps } from "@mui/material";
+import { styled } from "@mui/material/styles";
+import { FunctionComponent } from "react";
+
+const StyledGridCard = styled(Card)({
```
HenryQuillin, line 9: It looks like this component is a bit larger than the previous song card component.

HenryQuillin (approved):

src/pages/home/Home.tsx
scandycuz: It looks like container is just designed to be the basic center aligned container element for layouts.

src/components/SquareGridCard.tsx
scandycuz: Yeah, the component doesn't have a fixed width, that way the grid items don't overlap on smaller screens. They just take up a percentage of the total width of their container. We can decrease the containe width so that they closer match the mock-ups.

src/pages/home/Home.tsx
scandycuz: I think the main difference is that you can use the theme assigned breakpoints ("sm", "md", etc...) instead of fixed pixel widths.

PR #30, opened 2022-02-25
CU-1wuc2pk_When-there-are-no-songs-show-lets-get-this-party-started-and-Image-Upload-Component_Mateo-Sanchez

mallow43:
The Component works and I commented out the sample data so that yall could see and review. Dont merge it so I can uncomment the sample data before we merge into master.

Things to note:
I did not add actual functionality to the upload form and it is just there for show because Henry told me that there would probably be a major style redesign. Besides that I split the Songs.tsx file into two. SongContainer which has the popup and then separated the grid component into a SongGrid file which includes all the new code. Let me know if yall have any other questions or want me to add upload form functionality.


mallow43:
Task linked: [CU-1wuc2pk When there are no songs, show ‘lets get this party started’ and Image Upload Component.](https://app.clickup.com/t/1wuc2pk)

src/modules/song/slice.ts
```diff
@@ -1,238 +1,238 @@
...
+    //   description: "Description",
+    //   albumImage:
+    //     "https://images.complex.com/complex/images/c_fill,dpr_auto,f_auto,q_90,w_1400/fl_lossy,pg_1/hcjrqlvc6dfhpjxob9nt/cudi",
+    //   contributors: {
+    //     1: { name: "John", role: Role.Producer, stake: 0.25 },
+    //     2: { name: "Dan", role: Role.SoundEngineer, stake: 0.25 },
+    //     3: { name: "Kid Cudi", role: Role.Singer, stake: 0.5 },
+    //   },
+    //   extraInformation: "extra info",
+    // },
```
scandycuz, line 235: In this case, I think it would be fine to leave the mock data as is, and then just have instructions in the PR description that the reviewer should comment out the code in order to view the changes.

src/pages/home/songs/SongGrid.tsx
```diff
@@ -0,0 +1,104 @@
...
+export interface SongsProps {
+  history: History;
+  setOpenPopup: Dispatch<SetStateAction<boolean>>;
+}
+
+const Songs = ({ history, setOpenPopup }: SongsProps) => {
+    const songs = useAppSelector(selectSongs);
+  if (songs[0] === undefined) {
+    return (
+      <div style={ { height: "100%", width: "100%" } }>
```
scandycuz, line 20: `songs` will actually be an object, so `songs[0]` will only check if a song with an ID of 0 exists. You'd need to do something like 

```
const isSongsEmpty = Object.values(songs).length === 0;
```

Also, in order to clean up the component, this coud be moved to a separate component and imported. E.g.

```
import StartParty from "./StartParty"

...
if (isSongsEmpty) {
  return <StartParty />;
}

...
```

src/pages/home/songs/SongContainer.tsx
```diff
@@ -0,0 +1,36 @@
...
+import Popup from "components/Popup";
+import { History } from "history";
+import { useState } from "react";
+import SongGrid from "./SongGrid";
+import SongUploadForm from "./SongUploadForm";
+export interface SongsProps {
+  history: History;
+}
+
+const Songs = ({ history }: SongsProps) => {
```
scandycuz, line 11: Idealy, the file name and component name should be the same, just to keep it obvious what component is being exported.

I'm also hesitant to call it `SongContainer`, as historically a `Container` in React refers to a component that connects the app state to another component. A new developer may see the component name and assume that is what it is.

With functional components and Redux, Redux suggests using hooks to reference the app state in multiple components, instead of having a single container. I think just keeping the name the same as the page (`Songs`) probably works better.

src/components/styled/PinkTypography.ts
```diff
@@ -0,0 +1,10 @@
+import { Typography } from "@mui/material";
+import { styled } from "@mui/material/styles";
+
+const StyledTypography = styled(Typography)(({ theme }) => ({
+  color: theme.palette.primary.main,
+  fontWeight: 900,
+  padding: "34px 0 40px 0",
+  width: "100%",
+}));
+export default StyledTypography;
```
scandycuz, line 10: I'm not sure this necessarilly needs to be it's own component. I think we may end up having too many components if we create one for each text variation in the app.

It would be better to update the theme with the styling for an `h1` tag, reference that, and then use the standard MUI Typography component. E.g.

```
// theme.ts

export default createTheme({
  ...
  typography: {
  ...
  h1: {
    fontFamily: "Raleway",
    fontSize: "34px",
    fontWeight: 900,
  })
  ...
})

```

```
...
// Songs.tsx

  ...
  <Box pt="34px" pb="40px">
    <Typography variant="h3" color="primary" align="center" /> 
  </Box>
  ...
```

scandycuz (commented):
Looks good, but I had some comments

src/components/styled/PinkTypography.ts
mallow43: We might not want to apply these styles to every Typography h1. Would it be better to just make the styles inline or what do you think. For now Ill just commit them inline but let me know if you want me to change that.

scandycuz: I'm fine with making them inline. It looks like in this case, we're actualy just adding some custom styles to the existing `h5` styles, which I think is completely valid.

scandycuz (approved):

PR #34, opened 2022-02-26
Added Wallet Overview UI  

HenryQuillin:


src/components/styled/AnimatedGradientLine.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const AnimatedGradient = keyframes`
+    0%{background-position:0% 50%}
+    50%{background-position:100% 50%}
+    100%{background-position:0% 50%}
+`;
+
+const AnimatedGradientLine = styled("hr")({
+  animation: `${AnimatedGradient} 10s ease infinite;`,
+  background: "linear-gradient(270deg, #cc33cc, #323290);",
```
scandycuz, line 12: Ideally, we should be referencing theme values here. That way we ensure colors are consistent throughout the app.

scandycuz (approved):
Had one comment, but looks good to me

j-mateo (approved):

PR #37, opened 2022-03-03
Add song overview list view henry quillin

HenryQuillin:
## What

- Adds song list view on small screens

### Also

- Adds subtle animation to the gradient line above the list view 
- Removes unnecessary prop drilling for history. Adds the useHistory() hook to the components instead. 

## Test

1. Decrease your screen size to less than 900 pixels in width
2. Confirm that playlists are displayed in a list
3. Confirm that clicking appends the songId to history

### DIFF
--- a/src/components/TableLine.tsx
+++ b/src/components/TableLine.tsx
@@ -3,7 +3,7 @@
  */
 
 import { FunctionComponent } from "react";
-import HorizontalLine from "./styled/HorizontalLine";
+import StyledAnimatedGradientLine from "./styled/AnimatedGradientLine";
 
 interface TableLineProps {
   columns: number;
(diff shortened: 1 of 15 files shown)


j-mateo:
Task linked: [CU-24525km Song Overview List View](https://app.clickup.com/t/24525km)

src/components/TableLine.tsx
```diff
@@ -3,7 +3,7 @@
  */
 
 import { FunctionComponent } from "react";
-import HorizontalLine from "./styled/HorizontalLine";
```
scandycuz, line 6: I'd like to discuss this animation with the design team. I think we should only be adding animations where we are intentionally trying to draw the user's attention to that element.

scandycuz (approved):
Looks good to me, just had one comment.

src/components/TableLine.tsx
HenryQuillin: Ok I agree. Let's discuss with design and change if needed.

scandycuz: Sounds good, thanks.

HenryQuillin: For the time being I'll make it so slow that its barely noticable

scandycuz: Ok

PR #38, opened 2022-03-05
Home screen tab buttons

scandycuz:
## What

- Implements home screen tab buttons to switch between list and grid views

### Also

- Makes some minor refactors

### DIFF
--- a/src/pages/home/Home.tsx
+++ b/src/pages/home/Home.tsx
@@ -96,36 +96,17 @@ const Home: FunctionComponent<HomePropTypes> = (props) => {
           </TabPanel>
           <TabPanel value={ value } page={ Page.playlists }>
             <Playlists />
-<<<<<<< Updated upstream
-          </Container>
-        </TabPanel>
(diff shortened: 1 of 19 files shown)


j-mateo:
Task linked: [CU-2f70y29 Songs / playlists grid and list view buttons](https://app.clickup.com/t/2f70y29)

src/pages/home/Home.tsx
```diff
@@ -30,64 +38,77 @@ const Home: FunctionComponent<HomePropTypes> = (props) => {
     setValue(newValue);
   };
 
+  const isSmallScreen = useMediaQuery(theme.breakpoints.down("md"));
```
j-mateo, line 41: Do you think it makes sense to pull this out as an utility so it can be reused? Seems like something we may want to check in multiple components

j-mateo (approved):
lgtm!

src/pages/home/Home.tsx
scandycuz: Yeah, I think that would be helpful

PR #39, opened 2022-03-06
CU-24vyq5d_Fix-issues-with-prettier_Mateo-Sanchez

mallow43:
Created prettierrc.json and removed alphabetization of import keys from eslintrc.js

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -45,10 +45,6 @@ module.exports = {
     "import/order": [
       "warn",
       {
-        // alphabetize: {
-        //   caseInsensitive: true,
-        //   order: "asc",
-        // },
         groups: ["builtin", "external", "index", "sibling", "parent", "internal"],
       },
     ],
@@ -93,15 +89,6 @@ module.exports = {
         ignoreMemberSort: false,
       },
     ],
-    // "sort-keys": [
-    //   "warn",
-    //   "asc",
-    //   {
-    //     caseSensitive: true,
-    //     minKeys: 2,
-    //     natural: false,
-    //   },
-    // ],
   },
   settings: {
     react: {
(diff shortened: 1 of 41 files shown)


mallow43:
Task linked: [CU-24vyq5d Fix issues with prettier](https://app.clickup.com/t/24vyq5d)

.prettierrc.json
```diff
@@ -0,0 +1,6 @@
+{
+    "semi": true,
+    "trailingComma": "none",
```
j-mateo, line 3: just out of curiosity, why did we decide against trailing commas?

j-mateo (approved):
lgtm, just a few questions
Also, are you planning to put up another PR to apply the new format to existing files, or just let them get fixed organically as they get edited?

.prettierrc.json
HenryQuillin: Not sure. Personally, I'm all for trailing commas. It makes adding properties much easier.

scandycuz: I'm in favor of trailing commas as well

.eslintrc.js
```diff
@@ -45,10 +45,10 @@ module.exports = {
     "import/order": [
       "warn",
       {
-        alphabetize: {
-          caseInsensitive: true,
-          order: "asc",
-        },
+        // alphabetize: {
```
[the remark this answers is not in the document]
scandycuz: I agree we should just remove this code instead of commenting it out

scandycuz (dismissed):
I think we should make the suggested changes before merging it.

mallow43:
@j-mateo I was having trouble replying to your comment. But yeah I was just hoping to have it happen organically because if not I was thinking everyones working code will have a ton of merge conflicts. But maybe there is actually a way to do it so let me know.

PR #42, opened 2022-03-24
Implements RTK Query

scandycuz:
### What

Implements RTK Query in the app for songs (this still won't work as the authorization isn't set up, but this initial implementation should provide a starter template for implementing other API calls such as registration)

### Note

Since this doesn't actually add any app functionality, it shouldn't need to be tested

### DIFF
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -1,0 +1,11 @@
+import { createApi, fetchBaseQuery } from "@reduxjs/toolkit/query/react";
+
+const api = createApi({
+  reducerPath: "api",
+  baseQuery: fetchBaseQuery({
+    baseUrl: "https://staging-newm-server.herokuapp.com/v1/",
+  }),
+  endpoints: () => ({}),
+});
+
+export default api;
--- a/src/modules/song/api.ts
+++ b/src/modules/song/api.ts
@@ -6,7 +6,6 @@ const extendedApi = api.injectEndpoints({
       query: () => "portal/songs",
     }),
   }),
-  overrideExisting: false,
 });
 
 export default extendedApi;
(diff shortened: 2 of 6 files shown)


j-mateo:
Task linked: [CU-2jku1e3 Implement RTK Query (to enable making API calls)](https://app.clickup.com/t/2jku1e3)

src/api/api.ts
```diff
@@ -0,0 +1,11 @@
+import { createApi, fetchBaseQuery } from "@reduxjs/toolkit/query/react";
+
+const api = createApi({
+  reducerPath: "api",
+  baseQuery: fetchBaseQuery({
+    baseUrl: "https://staging-newm-server.herokuapp.com/v1/",
+  }),
+  endpoints: () => ({}),
+});
+
```
scandycuz, line 10: This sets up the base url for the NEWM API

src/modules/song/api.ts
```diff
@@ -0,0 +1,12 @@
...
+
+const extendedApi = api.injectEndpoints({
+  endpoints: (build) => ({
+    getSongs: build.query({
+      query: () => "portal/songs",
+    }),
+  }),
+  overrideExisting: false,
+});
+
```
scandycuz, line 11: Individual endpoints can be added in each module by extending the base API

src/modules/song/slice.ts
```diff
@@ -1,264 +1,27 @@
...
   name: "role",
   reducers: {},
+  extraReducers: (builder) => {
+    builder.addMatcher(
+      api.endpoints.getSongs.matchFulfilled,
+      (state, { payload }) => {
+        state.songs = payload.songs;
+      }
+    );
+  },
```
scandycuz, line 24: These matchers will update the Redux store when specific queries are completed

HenryQuillin (approved):
LGTM

PR #44, opened 2022-03-31
Adds social logins to sign-up screen

scandycuz:
### Testing instructions

Because the dynamically generated url for the PR is not an allowed origin for the social logins, you will need to pull down the branch and test it locally at http://localhost:3000.

You will also need to reach out to either Walter Lara or I to get the Google, Facebook, and LinkedIn client IDs and add them to your local `.env` file.

- To test the Google login, you will need to use your NEWM Gmail account
- To test the Facebook login, you will have to reach out to Walter Lara so that he can add your Facebook account as a test user
- To test the LinkedIn login, you can use any LinkedIn account

### Note

The API is restricted to one user account per email, so if you want to sign up with a social account that has the same email as one you've already signed up with, you'll need to first delete your existing account by making a DELETE request to the https://staging-newm-server.herokuapp.com/v1/users/me endpoint (using Postman or a curl request) with your session JWT as the authorization bearer token header

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -31,6 +31,7 @@ module.exports = {
   ],
   root: true, // For configuration cascading.
   rules: {
+    "testing-library/prefer-screen-queries": 0,
     "comma-dangle": [
       "error",
       {


j-mateo:
Task linked: [CU-2gmtnc9 Signup with social login](https://app.clickup.com/t/2gmtnc9)

HenryQuillin (approved):
Just one small change to remove a listing rule that was appearing in all the tests

PR #55, opened 2022-04-07
Sets up form and routes for "create profile" steps

scandycuz:
Screen Shot 2022-04-07 at 12 08 12 PM

### DIFF
--- a/src/pages/createProfile/CreateProfile.tsx
+++ b/src/pages/createProfile/CreateProfile.tsx
@@ -1,6 +1,6 @@
 import { FunctionComponent } from "react";
 import { Form, Formik } from "formik";
-import { Route, useRouteMatch } from "react-router-dom";
+import { Redirect, Route, Switch, useRouteMatch } from "react-router-dom";
 import * as Yup from "yup";
 
 interface FormValues {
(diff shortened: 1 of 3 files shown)


j-mateo:
Task linked: [CU-2pq6vfb Set up sign-up form context and routes](https://app.clickup.com/t/2pq6vfb)

HenryQuillin (commented):
Can we disable the authentication while we are still in development (or at least until we implement persistent user state)? This way we wouldn't have to add the API keys for every PR and branch, and we wouldn't have to sign in after every reload

src/pages/createProfile/CreateProfile.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  return (
+    <Formik
+      initialValues={ initialValues }
+      validationSchema={ ValidationSchema }
+      onSubmit={ handleSubmit }
+    >
+      { () => (
+        <Form>
+          <Switch>
```
HenryQuillin, line 33: Do we want to upgrade react-router-dom to V6? if so, we should do it now while the routing in our project is still pretty light. If we do, this would become ['Routes' instead of 'Switch'](https://reactrouter.com/docs/en/v6/upgrading/v5#upgrade-all-switch-elements-to-routes).

HenryQuillin (approved):

HenryQuillin:
> Can we disable the authentication while we are still in development (or at least until we implement persistent user state)? This way we wouldn't have to add the API keys for every PR and branch, and we wouldn't have to sign in after every reload
 Never mind just now seeing your slack message

src/pages/createProfile/CreateProfile.tsx
scandycuz: Sounds good to me. We can make that upgrade in a separate PR

scandycuz:
It might be good to prioritize adding temporary login and logout buttons in the UI (maybe absoutely positioned in the upper right hand corner) for until the actual UI is ready. If we do that, then we can add the persist session functionality, and it will make it should make development and testing easier.

PR #56, opened 2022-04-09
Update react-router-dom from v5 to v6

HenryQuillin:
Updates react router-dom v5 to react-router dom v6


j-mateo:
Task linked: [CU-2ba1nr8 Update react-router-dom to V6 and refactor](https://app.clickup.com/t/2ba1nr8)

src/App.tsx
```diff
@@ -20,27 +20,31 @@ const App = () => {
...
-            </Switch>
+            <Routes>
+              <Route path="/" element={ <Navigate to="home" replace /> } />
+
+              <Route path="linkedin" element={ LinkedInCallback } />
+
+              <Route path="login" element={ <Login /> } />
+
+              <Route path="sign-up" element={ <SignUp /> } />
+              <Route
```
HenryQuillin, line 31: We have to wrap PrivateRoutes in a Route component in v6, as only <Route>s can be children of <Routes>

src/common/hooks.ts
```diff
@@ -45,15 +45,17 @@ export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector;
...
-  const location = useLocation<{ from?: string }>();
-
+  const navigate = useNavigate();
+  const location = useLocation();
   useEffect(() => {
     if (isLoggedIn) {
-      history.push(location.state?.from || "/home");
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      location?.state?.from || navigate("/home");
```
HenryQuillin, line 58: So apparently in V6 we cannot add any types to useLocation. ( I was getting this error: `Expected 0 type arguments, but got 1`).

src/components/PrivateRoute.tsx
```diff
@@ -1,28 +1,17 @@
...
 import { selectSession } from "modules/session";
-import { Redirect, Route, RouteProps } from "react-router-dom";
+import { Navigate } from "react-router-dom";
 
-const PrivateRoute: FunctionComponent<RouteProps> = ({ children, ...rest }) => {
+type PrivateRouteProps = {
+  children: JSX.Element;
+};
+
+const PrivateRoute: FunctionComponent<PrivateRouteProps> = ({ children }) => {
```
HenryQuillin, line 10: Because we will now always be wrapping `<PrivateRoute>` components in a `<Route>` component, there is no need for rendering a `<Route>` component

src/pages/createProfile/CreateProfile.tsx
```diff
@@ -30,26 +29,29 @@ const CreateProfile: FunctionComponent = () => {
...
         <Form>
-          <Switch>
-            <Redirect
-              exact
-              to={ `${path}/what-should-we-call-you` }
-              from={ `${path}` }
+          <Routes>
+            <Route
+              path=""
+              element={ <Navigate to={ "what-should-we-call-you" } replace /> }
```
HenryQuillin, line 35: full path is `/create-profile/what-should-we-call-you`

src/common/hooks.ts
scandycuz: Looks good, but

src/components/PrivateRoute.tsx
```diff
@@ -1,27 +1,20 @@
...
-              state: { from: location },
-            } }
-          />
-        )
-      }
-    />
+  return isLoggedIn ? (
+    children
+  ) : (
+    <Navigate to={ "/sign-up" } state={ { from: location } } replace />
```
scandycuz, line 17: I haven't tested it, but I think something like this might be cleaner?

```
return (
  <Route 
    {...props} 
    element={(
      isLoggedIn
        ? element
        : <Navigate to="/sign-up" />
    )}
  />
)

```

Then, in the routes, you could just use

```
  <PrivateRoute 
    path="home" 
    element={<Home />}
  />

```

src/modules/session/slice.ts
```diff
@@ -8,7 +8,7 @@ interface SessionState {
 }
 
 const initialState: SessionState = {
-  isLoggedIn: false,
+  isLoggedIn: true, // set to true for development and tesing
```
scandycuz, line 11: I think you forgot to change this back

scandycuz (changes requested):
Looks good, but had some comments

src/common/hooks.ts
scandycuz, line 58: Looks good, but `location?.state?.from` still needs to be navigated to if it's present.

scandycuz (changes requested):

src/common/hooks.ts
HenryQuillin: Oops haha yea Idk how I missed that 🤦‍♂️

src/modules/session/slice.ts
HenryQuillin: I was thinking that this should stay true until we get persistent authentication. What do you think?

scandycuz: Fair point, that probably makes sense

src/components/PrivateRoute.tsx
HenryQuillin: Only a `<Route>` component can be a child of a `<Routes>` component. This means we can't use the`<PrivateRoute>` component as a direct child of `<Routes>`, which is why I had to change the current implementation. 

From the docs: 

"
[Refactor custom Routes
](https://reactrouter.com/docs/en/v6/upgrading/v5#refactor-custom-routes)Replace any elements inside a `<Switch>` that are not plain `<Route>` elements with a regular <Route>. This includes any `<PrivateRoute>`-style custom components.

You can [read more about the rationale behind this here](https://gist.github.com/mjackson/d54b40a094277b7afdd6b81f51a0393f), including some tips about how to use a `<Route render> `prop in v5 to achieve the same effect.
"

scandycuz: Technically, it should still be a Route component. Does it enforce the component having to have the name "Route"?

scandycuz: Read through the docs. Seems silly of them to not let you wrap a Route in your own component just because they thought people wouldn't like having to define the props.

scandycuz (approved):

src/components/PrivateRoute.tsx
HenryQuillin: I agree. Kinda strange

PR #59, opened 2022-04-16
Adds persistent session and refresh token functionality

scandycuz:
- Adds persistent session and refresh token functionality

### Also

- Adds temporary buttons to enable logging in and logging out until final UI is completed
- Updates api url paths to match recent updates to API

### Refactors

- Moves slice initial state type definitions into "types.ts" files


j-mateo:
Task linked: [CU-2pq8c9a Temporary login / logout buttons](https://app.clickup.com/t/2pq8c9a)

src/modules/session/slice.ts
```diff
@@ -19,14 +18,10 @@ const sessionSlice = createSlice({
     setSessionErrorMessage(state, action: PayloadAction<string>) {
       state.errorMessage = action.payload;
     },
+    logOut: handleLogout,
```
HenryQuillin, line 21: Is the state automatically getting passed into handleLogout? It could be written as 
`
    logOut(state) {
      handleLogout(state);
    },
`  correct? Just curious

HenryQuillin (commented):
LGTM. I like the decision to move the module types to their own files

escobarjonatan (approved):

src/modules/session/slice.ts
scandycuz: Yeah, that should be equivilent. It that notation, your more explicity calling the function, and in the current notatin, your just setting that object value to the function.

Similarly, `setSessionErrorMessage` could also be written as 

```
setSessionErrorMessage: (state, action: PayloadAction<string>) => {
  state.errorMessage = action.payload;
},
```

PR #60, opened 2022-04-17
Introduce a new dropdown select component and ran prettier for consistency

escobarjonatan:
MOCKS:
image

Storybook:
image


j-mateo:
Task linked: [CU-2cxepj3 Create Dropdown input component](https://app.clickup.com/t/2cxepj3)

src/App.css
```diff
@@ -9,7 +9,7 @@
 }
 
 .MuiSvgIcon-root {
-  color: #cc33cc !important;
+  color: #cc33cc;
```
escobarjonatan, line 12: I tried to find the reason for the need for `!important` but couldn't find anything. This was causing the icons to be this color. If someone knows we can try to find a solution that works for all

src/elements/DropdownSelect.tsx
```diff
@@ -0,0 +1,131 @@
...
+}
+
+const StyledDropdownSelectContainer = styled.div`
+  position: relative;
+`;
+
+const StyledResultsList = styled.ul`
+  background-color: ${theme.colors.grey500};
+  border-radius: 4px;
+  border: 2px solid ${theme.colors.grey400};
```
escobarjonatan, line 26: Went with 2px hardcoded over variable in case inputs value changes

src/elements/DropdownSelect.tsx
```diff
@@ -0,0 +1,131 @@
...
+          placeholder={ placeholder }
+          required={ required }
+          onKeyDown={ preventFormSubmit }
+        />
+      </div>
+
+      { hasResults ? (
+        <StyledResultsList { ...getListboxProps() }>
+          { (groupedOptions as typeof options).map((option, index) => (
+            <li { ...getOptionProps({ option, index }) } key={ option.id }>
```
escobarjonatan, line 117: Current valid option structure:
`{ id: 1, name: "Alternative", value: "Alternative" }`
The `id` property can be eliminated if needed, but if possible we should try to provide an `id` instead of using the map's `index` for rerendering/performance purposes. Let me know if you feel like we should use the map's `index` as a `key` instead.

src/elements/TextInput.tsx
```diff
@@ -33,10 +33,7 @@ const StyledInputElement = styled.input`
   font-size: ${theme.inputField.fontSize};
   font-weight: ${theme.inputField.fontWeight};
   line-height: ${theme.inputField.lineHeight};
-
-  input {
-    padding: 0;
-  }
+  padding: ${theme.inputField.padding};
```
escobarjonatan, line 36: I moved the padding to here so we can increase the clickable area for an input to be interacted with.

scandycuz: Nice, definitely a good idea.

src/App.css
scandycuz: I think that should be fine, always good to avoid the `!important` flag when possible.

src/elements/DropdownSelect.tsx
scandycuz: Sounds good to me, I know the React docs recommend against using the index as the key.

scandycuz (commented):
Looks great! I can take a more in-depth look this evening!

scandycuz (approved):

PR #61, opened 2022-04-19
Hotfix: Text input ref issue

scandycuz:
Fixes an error in the console due to the text input ref

### Also

- Minor updates

### DIFF
--- a/src/elements/DropdownSelect.tsx
+++ b/src/elements/DropdownSelect.tsx
@@ -1,4 +1,4 @@
-import { FunctionComponent, KeyboardEvent, MouseEvent } from "react";
+import { FunctionComponent, KeyboardEvent } from "react";
 import { useAutocomplete } from "@mui/base/AutocompleteUnstyled";
 import ArrowDropDownIcon from "@mui/icons-material/ArrowDropDown";
 import styled from "styled-components";
--- a/src/elements/DropdownSelect.tsx
+++ b/src/elements/DropdownSelect.tsx
@@ -1,4 +1,4 @@
-import { FunctionComponent, KeyboardEvent } from "react";
+import { FunctionComponent, KeyboardEvent, MouseEvent } from "react";
 import { useAutocomplete } from "@mui/base/AutocompleteUnstyled";
 import ArrowDropDownIcon from "@mui/icons-material/ArrowDropDown";
 import styled from "styled-components";
@@ -11,9 +11,8 @@ export interface SelectProps {
   readonly label: string;
   readonly name: string;
   readonly noResultsText?: string;
-  options: { id: number; name: string; value: string }[];
+  readonly options: { id: number; name: string; value: string }[];
   readonly placeholder?: string;
-  readonly required?: boolean;
 }
 
 const StyledDropdownSelectContainer = styled.div`
@@ -61,7 +60,6 @@ const DropdownSelect: FunctionComponent<SelectProps> = ({
   noResultsText = "Nothing found",
   options,
   placeholder,
-  required,
 }) => {
   const {
     getInputProps,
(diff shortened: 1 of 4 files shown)


src/elements/DropdownSelect.tsx
```diff
@@ -75,23 +73,41 @@ const DropdownSelect: FunctionComponent<SelectProps> = ({
...
   const hasResults = groupedOptions.length > 0;
   const showNoResults = !hasResults && popupOpen;
 
-  // This prevents a form submission when input text does not match any options
+  const { onMouseDown, ...inputProps } = getInputProps();
+
+  /**
+   * This prevents a form submission when input
+   * text does not match any options.
+   */
```
scandycuz, line 87: I think using JSDOC comments to describe functions can look a bit nicer

src/elements/DropdownSelect.tsx
```diff
@@ -102,11 +118,11 @@ const DropdownSelect: FunctionComponent<SelectProps> = ({
               } }
             />
           }
+          onMouseDown={ handleMouseDown }
           errorMessage={ errorMessage }
           label={ label }
           name={ name }
           placeholder={ placeholder }
-          required={ required }
```
scandycuz, line 109: We're using our own validations, so I don't think we need this

src/elements/DropdownSelect.tsx
```diff
@@ -75,23 +73,41 @@ const DropdownSelect: FunctionComponent<SelectProps> = ({
...
+   */
   const preventFormSubmit = (event: KeyboardEvent): void => {
     if (event.key === "Enter" && inputValue !== value?.name)
       event.preventDefault();
   };
 
+  /**
+   * Disables autocomplete click handler if input is disabled.
+   */
+  const handleMouseDown = (event: MouseEvent<HTMLInputElement>) => {
```
escobarjonatan, line 96: I think that fixing the ref issue would solve this problem, so this might not be needed. I tested it on my end with a native input and it marked it as disabled, that's why i didn't add logic for the disabled state

src/elements/TextInput.tsx
```diff
@@ -2,25 +2,32 @@ import {
...
+  readonly startAdornment?: JSX.Element;
+  readonly endAdornment?: JSX.Element;
 }
 
 const StyledRootElement = styled.div`
   display: flex;
   justify-content: space-between;
   align-items: center;
   flex-grow: 1;
+  padding: ${theme.inputField.padding};
```
escobarjonatan, line 27: I noticed that this var is set to `9px 13px` which is what figma has, but it doesn't go well with the 4px grid system and in the long run it will throw things off by a pixel or 2. I was going to suggest we change that value to `8px 12px`. What do you think?

escobarjonatan: Also, do we want to tie the styles of the root element to the `theme.inputField.padding`? 
I see we are adding padding in the root and in the input, we should only be adding it to one right? 
I might be missing something here though 🤔

scandycuz: For the pixels, I was just going by the dimensions in the Figma mock-up, but we could change it.

For the padding, unfortunately, the start and end adornments look incorrect if we only add padding to the input, since the padding needs to be outside of them. I'm realizing that I need to do the same thing for the left padding though, so tha the "start" adornment will show up correctly.

escobarjonatan: Let's check with design and understand where they were going with this. I dont think this is something that needs to be fixed with this PR, but worth a discussion.

src/elements/DropdownSelect.tsx
scandycuz: Good call, it looks like the MUI `useInput` wasn't respecting the disabled prop.

escobarjonatan (approved):

PR #62, opened 2022-04-19
DropdownSelect | Adding unit tests and fixing the use of theme in typography

escobarjonatan:


j-mateo:
Task linked: [CU-2egzdc3 Dropdown | add unit tests](https://app.clickup.com/t/2egzdc3)

src/elements/Typography.tsx
```diff
@@ -20,8 +20,6 @@ const Typography: FunctionComponent<TypographyProps> = ({
   sx,
   ...rest
 }) => {
-  const theme = useTheme();
```
escobarjonatan, line 23: This was throwing an error in the console and i think that we are using the theme from `src/theme/theme.ts`. @scandycuz let me know if i should put it back and continue using the `useTheme` from MUI

scandycuz: They should reference the same theme. Was it throwing an error in the test? The component likely just needs to be wrapped in the theme provider for the hook to work. If you use the `renderWithContext` function from the `common/testUtils` file, instead of the `render` function from the React testing library, it should resolve the error.

src/elements/test/DropdownSelect.test.tsx
```diff
@@ -0,0 +1,125 @@
+import { render, screen } from "@testing-library/react";
```
scandycuz, line 1: Could use the `renderWithContext` function here instead of `render` to make the `useTheme` hook work.

scandycuz (approved):
LGTM!

src/elements/Typography.tsx
escobarjonatan: I reverted the change and ran the tests again and they seemed to run fine. There are other issues with our tests though that we can look into later on

src/elements/test/DropdownSelect.test.tsx
escobarjonatan: wait a minute... I think i know what you where you were going with that ^, i will change it and run the tests again

scandycuz: Yeah, I thought about splitting them into a Redux context and theme context function, but ended up just combining them into a single function.

src/elements/test/DropdownSelect.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+      errorMessage: undefined,
+      label: "Label",
+      name: "name",
+      noResultsText: undefined,
+      options: data,
+      placeholder: undefined,
+      ...propOverrides,
+    };
+
+    return await renderWithContext(<DropdownSelect { ...props } />);
```
scandycuz, line 30: What was the reason for using an async / await function?

scandycuz (approved):

PR #67, opened 2022-04-27
WIP: Implementing create account flow UI for new users

escobarjonatan:
Note this is still WIP as we are waiting on design to make a decision on small details

Welcome page mock: 
image

Welcome implementation:
image

Verification page mock:
image

Verification implementation:
image

### DIFF
--- a/src/modules/session/api.ts
+++ b/src/modules/session/api.ts
@@ -1,11 +1,13 @@
 import api from "api";
 import { EmptyResponse } from "common";
 import {
+  CreateAccountRequest,
   GetProfileResponse,
   LoginRequest,
   NewmAuthResponse,
   NewmOAuthRequest,
-  UpdateProfileRequest,
+  Request2FACode,
+  UpdateProfileRequest
 } from "./types";
 
 export const extendedApi = api.injectEndpoints({
@@ -62,6 +64,21 @@ export const extendedApi = api.injectEndpoints({
       query: (body) => ({
         url: "v1/users/me",
         method: "PATCH",
+        body,
+      }),
+    }),
+
+    sendVerificationEmail: build.query<EmptyResponse, Request2FACode>({
+      query: () => ({
+        url: "v1/auth/code",
+        method: "GET",
+      }),
+    }),
+
+    createAccount: build.mutation<EmptyResponse, CreateAccountRequest>({
+      query: (body) => ({
+        url: "v1/users",
+        method: "PUT",
         body,
       }),
     }),
--- a/src/pages/signUp/SignUp.tsx
+++ b/src/pages/signUp/SignUp.tsx
@@ -1,60 +1,107 @@
-import NEWMLogo from "assets/images/NEWMLogo";
-import { useAuthenticatedRedirect } from "common";
+import * as Yup from "yup";
+import { Box, Container } from "@mui/material";
+import { useTheme } from "@mui/material/styles";
+import { FormikValues } from "formik";
 import { FunctionComponent } from "react";
-import { Typography } from "elements";
-import { FacebookLogin, GoogleLogin, LinkedInLogin } from "components";
-import { Box, Container, Stack, useTheme } from "@mui/material";
+import { useDispatch } from "react-redux";
+import { WizardForm } from "components";
+import {
+  createAccount,
+  extendedApi as sessionApi
+} from "modules/session";
+import Verification from "./signUpSteps/Verification";
+import Welcome from "./signUpSteps/Welcome";
+
+interface AccountValues {
+  readonly authCode: string;
+  readonly confirmPassword: string;
+  readonly email: string;
+  readonly newPassword: string;
+}
 
 const SignUp: FunctionComponent = () => {
   const theme = useTheme();
+  const dispatch = useDispatch();
+  const initialValues: AccountValues = {
+    email: "",
+    newPassword: "",
+    confirmPassword: "",
+    authCode: "",
+  };
 
-  useAuthenticatedRedirect();
+  /**
+    * Password regex, it must contain the following:
+    * 8 characters, 1 uppercase letter, 1 lowercase letter, 1 number and one special character.
+  */
+  // TODO: Enable the regex when finished testing
+  // const passwordRegex = /^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[@$!%*?&])[A-Za-z\d@$!%*?&]{8,}$/;
+  const passwordRegex = /^([a-z])/;
 
-  return (
-    <Container
-      maxWidth="lg"
-      sx={ {
-        display: "flex",
-        minHeight: "100%",
-        backgroundColor: theme.palette.background.default,
-        justifyContent: "center",
-        alignItems: "center",
-      } }
-    >
-      <Box width="312px">
-        <Box
-          sx={ {
-            display: "flex",
-            flexDirection: "column",
-            flexGrow: 1,
-            justifyContent: "center",
-            alignItems: "stretch",
-          } }
-        >
-          <Box alignSelf="center">
-            <NEWMLogo />
-          </Box>
+  /**
+ * Yup validations for all form fields.
+ */
+  const validations = {
+    authCode: Yup.string().required(),
+    email: Yup.string().email("Please enter a vaild email").required("E-mail is required"),
+    newPassword: Yup.string().required("Password is required").matches(
+      passwordRegex,
+      "Minimum 8 characters, at least one uppercase letter, one lowercase letter, one number and one special character"
+    ),
+    confirmPassword: Yup.string()
+      .required("Confirm password is required").oneOf([Yup.ref("newPassword")], "Passwords must match"),
+  };
 
-          <Box mt={ 2 } display="flex" justifyContent="center">
-            <Typography align="center">Or sign up via</Typography>
-          </Box>
+  const sendVerificationEmail = (values: FormikValues) => {
+    dispatch(sessionApi.endpoints.sendVerificationEmail.initiate({ email: values.email }));
+  };
 
-          <Box
-            mt={ 1 }
-            display="flex"
-            justifyContent="center"
-            alignItems="center"
-          >
-            <Stack direction="row" spacing={ 2 }>
-              <GoogleLogin />
-              <FacebookLogin />
-              <LinkedInLogin />
-            </Stack>
-          </Box>
-        </Box>
+  /**
+   * Attempts to create an account on submit of the last form route.
+   */
+  const handleSubmit = ({ authCode, confirmPassword, email, newPassword }: FormikValues): void => {
+    dispatch(createAccount({ authCode, confirmPassword, email, newPassword }));
+  };
+
+    return (
+      <Box
+        sx={ {
+          backgroundColor: theme.colors.black,
+          display: "flex",
+          flex: 1,
+          maxWidth: "100%",
+          pt: 10,
+          px: 2,
+        } }
+      >
+        <Container maxWidth="xl">
+          <WizardForm
+            initialValues={ initialValues }
+            onSubmit={ handleSubmit }
+            rootPath="sign-up"
+            validateOnMount={ true }
+            routes={ [
+              {
+                element: <Welcome />,
+                onSubmitStep: sendVerificationEmail,
+                path: "",
+                validationSchema: Yup.object().shape({
+                  email: validations.email,
+                  newPassword: validations.newPassword,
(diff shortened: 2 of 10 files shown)


j-mateo:
Task linked: [CU-2p687p3 Registration steps "Web" 1.0 - 2.1](https://app.clickup.com/t/2p687p3)

src/modules/session/api.ts
```diff
@@ -65,6 +67,21 @@ export const extendedApi = api.injectEndpoints({
         body,
       }),
     }),
+
+    sendVerificationEmail: build.query<EmptyResponse, Request2FACode>({
+      query: () => ({
```
scandycuz, line 72: Looks good, but you'll also need to pass the `email` argument here. E.g:

```
query: ({ email }) => ({
  url: "v1/auth/code",
  method: "GET",
  params: { email },
}),
```
should work

scandycuz (commented):
Looks good, just needs one minor update for the email verification query to work.

src/pages/signUp/SignUp.tsx
```diff
@@ -1,59 +1,109 @@
...
+
+  /**
+ * Yup validations for all form fields.
+ */
+  const validations = {
+    authCode: Yup.string().required("Verification code is required"),
+    email: Yup.string().email("Please enter a vaild email").required("E-mail is required"),
+    newPassword: Yup.string().required("Password is required").matches(
+      passwordRegex,
+      "Minimum 8 characters, at least one uppercase letter, one lowercase letter, one number and one special character"
```
scandycuz, line 47: Is it possible to omit the special character requirement?

src/pages/signUp/signUpSteps/Welcome.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    return showPasswordIcon ? (
+      <IconButton
+        aria-label="Change password field visible status"
+        onClick={ togglePassword }
+      >
+        <PasswordIcon sx={ { color: theme.colors.white } } />
+      </IconButton>
+    ) : null;
+  };
```
scandycuz, line 45: I noticed that it was neccessary to click the `PasswordEndAdornment` icon twice in order for it to register the click. This is because the component is being re-rendered on each render, so it isn't retaining it's state correctly. If you move the `PasswordEndAdornment` component outside of the `SignUp` function, it will resolve the issue of requiring multiple clicks.

I think putting the `PasswordEndAdornment` component in the `components` directory would make sense, since it will likely be required on the login screen as well.

escobarjonatan: Good catch, i hadn't noticed that.

Maybe i can extract the whole password input and make that a component that has the toggle icon functionality in it too

scandycuz: That sounds like a good approach to me. Might be tricky since the password and confirm password inputs have to be in sync though. May have to handle the toggle icon functionality externally

src/pages/signUp/signUpSteps/Welcome.tsx
```diff
@@ -10,39 +7,27 @@ import {
...
-      >
-        <PasswordIcon sx={ { color: theme.colors.white } } />
-      </IconButton>
-    ) : null;
-  };
+  useEffect(() => {
+    newPassword || confirmPassword
+      ? setShowEndAdornment(true)
+      : setShowEndAdornment(false);
+  }, [newPassword, confirmPassword]);
```
scandycuz, line 30: Would it be more straightforwad to just use 

```
const showEndAdornment = !!(newPasword || confirmPassword)
```
instead of using `useState` and `useEffect`?

src/components/form/PasswordInputField.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { TextInputField } from "components";
+
+interface PasswordInputFieldProps {
+  readonly ariaLabel?: string;
+  readonly endAdornmentHandler?: () => void;
+  readonly externalMaskPassword?: boolean;
+  readonly name: string;
+  readonly placeholder?: string;
+  readonly showEndAdornment?: boolean;
+}
```
scandycuz, line 15: I'd probably just extend the `TextInputProps`, that way you don't have to define props such as the `ariaLabel` (I think it's also fine to not worry about setting a default for it). That way it will have all of the props of a normal text input defined, in addtion to the password field props. E.g.

```
import { TextInputProps } from "elements";

...

interface PasswordInputFieldProps extends TextInputProps {
  readonly handlePressEndAdornment?: VoidFunction; // just preference, either name works
  readonly externalMaskPassword?: boolean;
  readonly showEndAdornment?: boolean;
}
```

src/components/form/PasswordInputField.tsx
```diff
@@ -0,0 +1,58 @@
...
+  endAdornmentHandler,
+  externalMaskPassword = true,
+  name,
+  placeholder = "Password",
+  showEndAdornment = true,
+}) => {
+  const theme = useTheme();
+  const [internalMaskPassword, setInternalMaskPassword] = useState(true);
+
+  const isMasked = endAdornmentHandler ? externalMaskPassword : internalMaskPassword;
```
scandycuz, line 28: Might be more readable to use 

```
const isMasked = externalMaskPassword ?? internalMaskPassword
```

The nullish coalescing operator will only reference the second variable if the first variable is `null` or `undefined`

scandycuz (commented):
LGTM, just left a few comments

src/pages/signUp/test/Signup.test.tsx
```diff
@@ -3,10 +3,10 @@ import SignUp from "../SignUp";
...
   it("renders social login buttons", () => {
-    const { queryByLabelText } = renderWithContext(<SignUp />);
+    const { getByLabelText } = renderWithContext(<SignUp />);
 
-    expect(queryByLabelText("google authorization")).toBeTruthy();
-    expect(queryByLabelText("facebook authorization")).toBeTruthy();
-    expect(queryByLabelText("linkedin authorization")).toBeTruthy();
+    expect(getByLabelText("google authorization")).toBeTruthy();
+    expect(getByLabelText("facebook authorization")).toBeTruthy();
+    expect(getByLabelText("linkedin authorization")).toBeTruthy();
```
scandycuz, line 10: What was the reason for replacing `queryByLabelText` with `getByLabelText`? `queryByLabelText` should work better as it will just make the test fail if the text isn't present, where as `getByLabelText` will throw an error.

src/pages/signUp/signUpSteps/Verification.tsx
```diff
@@ -0,0 +1,71 @@
...
+          name="authCode"
+          placeholder="Verification Code"
+          type="text"
+        />
+        <FilledButton type="submit" disabled={ !isValid }>
+          Enter
+        </FilledButton>
+      </Stack>
+
+      <Box alignSelf="center" position="absolute" bottom="32px" left="0" right="0" color="grey200">
```
scandycuz, line 59: Looks like the absolute positioning is causing the text to be relative to the bottom of the screen, which may cause issues with the responsive layout.

Screen Shot 2022-05-03 at 4 34 28 PM

src/components/form/PasswordInputField.tsx
escobarjonatan: I thought about doing it this way but i saw an error in the console that i thought happened when spreading props:
image

I applied the changes though

src/components/form/PasswordInputField.tsx
escobarjonatan: This is part of the confusing stuff i mentioned to you...on this line: `const isMasked = endAdornmentHandler ? externalMaskPassword : internalMaskPassword;`  I am checking if there is a handler passed through... if so, we want to use the `externalMaskPassword` prop if none is provided, I am creating an internal masked state and toggling that value

src/pages/signUp/signUpSteps/Verification.tsx
escobarjonatan: So this is the reason why i asked for responsive font sizes... bc in the mock it looks like it's absolute positioned to the bottom which works fine for large screens but not so much for smaller screens. I think this shouldn't be an issue once we implement the responsive font sizes... Do you think i should take a different approach?

src/pages/signUp/signUpSteps/Welcome.tsx
escobarjonatan: Well i usually like to complicate things so no haha. Good catch, i was trying to do other logic inside the useEffect that is no longer applicable and forgot to clean this up

src/pages/signUp/test/Signup.test.tsx
escobarjonatan: Well this is interesting... my ts configs say we should be using get instead of query? 🤔
image

src/components/form/PasswordInputField.tsx
scandycuz: Interesting. Yeah, that may be the case. I still think extending the props is good though, that way the user can use all of the other text input props, such as `autoComplete`, etc...

src/components/form/PasswordInputField.tsx
scandycuz: Yeah, that works. I was thinking it might simplify the code if we just check the `externalMaskPassword` prop and use that to determine whether to use the internal value or not. The user probably wouldn't pass an `externalMaskPassword` prop unless they were also handling the end adornment funtionality externally, so I think it would be the same in regards to functionality. Either way sounds good to me though.

src/pages/signUp/test/Signup.test.tsx
scandycuz: Interesting, I didn't get that linting error. The libary documentation states that queryBy

> Returns the matching node for a query, and return null if no elements match. This is useful for asserting an element that is not present.

It definitely looks like the linter recommends `getBy` though. I believe that the issue with `getBy` is that the error it throws will cause the tests to end prematurely if the value isn't present, but all of the tests have to pass anyway, so I suppose either way works.

src/pages/signUp/signUpSteps/Verification.tsx
scandycuz: Hmm... well, if we're confident that no one's device will see it overlapping, either way sounds fine to me. Could do something like this to maintain it's presence at the bottom of the screen without absolute positioning it though:

```
<div style={{ minHeight="100vh" display: "flex"; flexDirection: "column", alignContent: "space-between">
  <div>
    // main content here
  </div>

  <div>
    // content at bottom of screen here
  </div>
</div>
``

**Review by scandycuz — APPROVED** [review]

**Review by scandycuz — COMMENTED** [review]
LGTM. Left some comments, but none of them are blocking. Just the password validation needs to be updated

File: src/components/form/PasswordInputField.tsx (hunk 6 above)
Comment (escobarjonatan) [reply src/components/form/PasswordInputField.tsx]: ohhh i see what you meant, the implementer of PasswordInputField would not pass externalMaskPassword without the handlePressEndAdornment... i tested it and it worked, i do agree it reads better that way. change coming in the next push

File: src/pages/signUp/test/Signup.test.tsx (hunk 7 above)
Comment (escobarjonatan) [reply src/pages/signUp/test/Signup.test.tsx:10]: I won't modify future tests that have the query but i think it's worth getting our linters in sync sometime

File: src/pages/signUp/signUpSteps/Verification.tsx (hunk 8 above)
Comment (escobarjonatan) [reply src/pages/signUp/signUpSteps/Verification.tsx:59]: Doing it this way would mean i need to reconsider the layout of the signup to not include the logo and go back to repeating the usage 🤔 I will keep this in mind when i do the responsive story... i think i will be making the change though

File: src/components/test/PasswordInputField.test.tsx [hunk 9]
```diff
@@ -0,0 +1,77 @@
...
+        onSubmit: jest.fn(),
+      })
+    );
+  };
+
+  it("renders with defaults when only required name prop is given", () => {
+    renderComponent();
+
+    expect(screen.getByTestId("passwordInput")).toBeInTheDocument();
+    expect(screen.getByTestId("VisibilityIcon")).toBeInTheDocument();
```
scandycuz, line 39: The React Testing libary [docs](https://testing-library.com/docs/queries/about/#priority) recommend using `getByTestId` only when another query won't work, since it doesn't match what the user sees. I'd recommend using the `getByPlaceholderText` query

escobarjonatan: good call, i can add a placeholder prop to it and use that

scandycuz (approved):

PR #73, opened 2022-05-26
Refactor typography styles

scandycuz:
## What

Refactors how typography is styled:

- Uses default Material UI variants (h1, h2, h3, body1, ect...). This will also set the HTML tag to the variant that is used.

Ideally, the variant can handle most of the styling for typography, and any additional styling can be added inline, either using one-off styles or theme values. E.g.

```
<Typography variant="h1" sx={theme.typography.emphasized}>
  Hello world
</Typography>
```

If we want to use a different HTML tag than the variant default, the Material UI Typography component allows for a `component` prop to specify it. E.g.

```
<Typography
  variant="h3"
  component="p"
>
  Hello world
</Typography>
```

### DIFF
--- a/src/elements/styled/GradientTypography.ts
+++ b/src/elements/styled/GradientTypography.ts
@@ -1,4 +1,4 @@
-import styled from "styled-components";
+import { styled } from "@mui/material/styles";
 import theme from "theme";
 import Typography from "../Typography";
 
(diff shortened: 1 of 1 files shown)


src/elements/GradientTextInput.tsx
```diff
@@ -69,6 +73,11 @@ const StyledInputElement = styled("input")<StyledInputElementProps>`
   &:focus {
     outline: none;
   }
+
+  ${mdBreakpoint} {
```
escobarjonatan, line 77: i see what you did here, nice! 🔥

escobarjonatan (approved):
LGTM!

PR #74, opened 2022-05-29
Updating profile page to accommodate account detail changes

escobarjonatan:
Implement the profile page for the users.

Looks:
image

[Figma Mocks](https://www.figma.com/file/zrn01ZR2zFeLiKBebEYI1v/Artist-Portal---Final-Designs?node-id=14%3A2989)

Notes:
- You'll notice that the first and last names are their own field, this is an intended change, see the discussion in figma
- User profile picture implementation will be worked in a separate PR


src/common/formUtils.ts
```diff
@@ -0,0 +1,46 @@
+import * as Yup from "yup";
+
+/**
+ * Password regex, it must contain the following:
+ * 8 characters, 1 uppercase letter, 1 lowercase letter and 1 number.
+ */
+const passwordRequirementRegex = /^(?=.*[0-9])(?=.*[a-z])(?=.*[A-Z]).{8,}$/;
+
+export const commonYupValidation = {
```
escobarjonatan, line 9: Since we are reusing these validations in multiple places and potentially more... i created this and we can expand on it, for example:

```
import { commonYupValidation } from "common";
...
  const validationSchema = Yup.object({
    firstName: commonYupValidation.firstName.matches(/^[aA-zZ\s]+$/, "Please only use letters"),
  });
...
```

src/common/mockConstants.tsx
```diff
@@ -10,7 +10,7 @@ export const mockProfile: Profile = {
   nickname: "Johnny",
   pictureUrl: "",
   role: "Producer",
-  genres: ["Electronic", "Instrumental"],
+  genre: "Instrumental",
```
escobarjonatan, line 13: Updated `genre` from the previous `genres` value

src/elements/TextInput.tsx
```diff
@@ -12,10 +12,11 @@ import theme from "theme";
 import Typography from "./Typography";
 
 export interface TextInputProps extends InputHTMLAttributes<HTMLInputElement> {
-  readonly label?: string;
+  readonly endAdornment?: JSX.Element;
   readonly errorMessage?: string;
+  readonly label?: string;
   readonly startAdornment?: JSX.Element;
-  readonly endAdornment?: JSX.Element;
+  readonly widthType?: "default" | "full";
```
escobarjonatan, line 19: Talked to design, they wanted to have 2 options for the text fields... either full width 100% or max width of `340px` to stay consistent throughout the app

src/pages/home/profile/Profile.tsx
```diff
@@ -1,17 +1,213 @@
...
+              <Stack
+                sx={ {
+                  display: "grid",
+                  gridTemplateColumns: [
+                    "repeat(1, 1fr)",
+                    null,
+                    "repeat(2, 1fr)",
+                  ],
+                  columnGap: [undefined, undefined, "20px"],
+                  maxWidth: [undefined, undefined, "700px"],
```
escobarjonatan, line 114: interesting find here: you'd think that null and undefined would give you similar results.... they do not.

null will insert a value that fits the media query of the screen and still show up... while undefined will not show a selector/property at all.

PR #75, opened 2022-06-04
2vxaken owners page UI henry

HenryQuillin:
Creates UI of the Owners Page 
Video Demo: https://imgur.com/hSzFyLo 
If there are no owners: 
Screen Shot 2022-06-03 at 10 40 28 PM

Note: This is just the UI. I did not create a Redux module because it seems like we are still figuring out what properties will be needed for the Owner type.


src/components/styled/StyledPagination.tsx
```diff
@@ -0,0 +1,37 @@
...
+  "& .MuiPaginationItem-previousNext": {
+    borderRadius: "0px 6px 6px 0px",
+  },
+  "& li:first-child > button": {
+    borderRadius: "6px 0px 0px 6px",
+  },
+
+  "& .MuiPaginationItem-ellipsis": {
+    paddingTop: "5px !important",
+    paddingBottom: "5px !important",
```
escobarjonatan, line 28: Is there a way that we can increase specificity to the selector so that we dont need to add the important here? 

Also, can these padding values be 4px instead of 5px? As we add more content to the pages, following the 4 point grid system will help with the vertical rhythm of the page and prevent alignment issues in the future

src/components/styled/StyledPagination.tsx
```diff
@@ -0,0 +1,37 @@
...
+  },
+
+  "& .Mui-selected": {
+    color: "white",
+    background: theme.colors.grey400,
```
escobarjonatan, line 33: In this case you are only changing the background color. It is good practice to use `background-color` in this case because using `background` property here will reset the rest of the properties background shorthand uses like repeat, position, color etc.  Unless you meant to reset the rest of the properties to their default, then you can keep it as is

src/components/styled/StyledPagination.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const grey100 = theme.colors.grey100;
+
+const StyledPagination = styled(Pagination)<PaginationProps>({
+  color: grey100,
+
+  "& .MuiPaginationItem-root": {
+    color: grey100,
+    borderRadius: "0px",
+    border: `1px solid ${grey100}`,
```
escobarjonatan, line 13: This value seems like a good candidate for the theme if we keep reusing it... no need to change it now, but let's keep our eyes on it

src/pages/home/owners/Owners.tsx
```diff
@@ -1,16 +1,76 @@
...
+      </Typography>
+
+      { ownersData.length ? (
+        <>
+          <Box sx={ { width: { xs: 200, sm: 380 }, pb: 3 } }>
+            <TextInput
+              value={ query }
+              onChange={ (e) => requestSearch(e.target.value) }
+              style={ {
+                width: "380px",
```
escobarjonatan, line 52: I talked to tomas and he mentioned having search inputs be max width of 340px to stay consistent throughout our app.... I believe you can set the width to 100% here and the text input will take care of the max width since it is defined in the text-input component
if not, you can use the `maxWidth` from the theme.inputField

src/pages/home/owners/Owners.tsx
```diff
@@ -1,16 +1,76 @@
...
+                  } }
+                />
+              }
+              placeholder="Search by owner or song"
+            ></TextInput>
+          </Box>
+          <OwnersTable ownersData={ filteredData } />
+        </>
+      ) : (
+        <Box sx={ { margin: "auto", position: "relative", bottom: "50px" } }>
```
escobarjonatan, line 70: Is the bottom 50px needed here? since this is a relative positioned element

src/pages/home/owners/OwnersTable.tsx
```diff
@@ -0,0 +1,136 @@
...
+    event: React.ChangeEvent<unknown>,
+    page: number
+  ) => {
+    setPage(page);
+  };
+
+  return (
+    <StyledTableContainer>
+      <Table sx={ { minWidth: 500 } } aria-label="Owners pagination table">
+        <TableHead sx={ { lineHeight: "16.94px" } }>
```
escobarjonatan, line 64: Curious here as to the specific line height value, the line height should be set  similar to our typography values

src/theme/theme.ts
```diff
@@ -226,8 +229,8 @@ const theme = createTheme({
       fontFamily: "Raleway",
       fontStyle: "normal",
       fontWeight: 800,
-      fontSize: "80px",
-      lineHeight: "96px",
+      fontSize: "32px",
+      lineHeight: "37.57px",
```
escobarjonatan, line 233: This is a big difference in values for all H1 variants.... now the h1 and h2 have the same size. Was this a change requested by design? 🤔

src/theme/theme.ts
```diff
@@ -264,6 +267,13 @@ const theme = createTheme({
       fontSize: "14px",
       lineHeight: "20px",
     },
+    body2: {
+      fontFamily: "Inter",
+      fontStyle: "normal",
+      fontWeight: 500,
```
escobarjonatan, line 273: The only difference i see between body1 and body2 is the fontWeight, i believe you can still use the body1 in your case and override it like this:

```
<Typography variant="body1" sx={theme.typography.fontWeightMedium}>
  Enter text here
</Typography>

```

escobarjonatan (approved):

src/components/styled/StyledPagination.tsx
HenryQuillin: I think I accidentally left those !important flags because it works fine without them. Also, good point about the padding, lets's just make sure that design is aware of the 4-point grid system.

src/components/styled/StyledPagination.tsx
HenryQuillin: I did mean to reset all the other properties 👍

src/pages/home/owners/Owners.tsx
HenryQuillin: Makes sense to me. I went ahead and changed it to 340px but the design specifies a width of 380

src/pages/home/owners/Owners.tsx
HenryQuillin: In the design, the message is centered in the middle of the page. Here, the relative container is pushed down by the OWNERS text, which is why the bottom 50px is necessary.

PR #76, opened 2022-06-04
Cu 2vx4wrm: upload song

scandycuz:
## What

- Implements upload song functionality

### Note

- Upload audio file and minting song will be implemented separately

## Look

Screen Shot 2022-06-04 at 3 02 17 AM

## Test

1. Upload a song using the form
2. Use the Redux logging to verify the song with the image url is fetched from the NEWM API


src/components/UploadImage.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+        const hasValidDimensions = await validateImageDimensions(
+          fileWithPreview.preview,
+          2048,
+          2048
+        );
+
+        if (!hasValidDimensions) {
+          throw new Error(
+            "Please upload an image with a height and width of least 2048 pixels."
```
HenryQuillin, line 73: Why does the file need to be at least 2048x2048px?

src/components/styled/DashedOutline.ts
```diff
@@ -0,0 +1,10 @@
+import { styled } from "@mui/material/styles";
+import { Box } from "@mui/material";
+import theme from "theme";
+
+const DashedOutline = styled(Box)`
```
HenryQuillin, line 5: Shouldn't this (and the ErrorMessage.ts and SolidOutline.ts) go under elements/styled instead of components/styled?

src/elements/Typography.tsx
```diff
@@ -1,23 +1,19 @@
...
-  useTheme,
 } from "@mui/material";
-import { Theme } from "@mui/material/styles";
+import { Theme, useTheme } from "@mui/material/styles";
 import { FunctionComponent } from "react";
 
 export interface TypographyProps extends MuiTypogaphyProps {
-  variant?: TypographyVariant;
   color?: keyof Theme["colors"];
   fontWeight?: "regular" | "medium" | "semi-bold" | "bold" | "extra-bold";
```
HenryQuillin, line 10: When you select text in Figma it does not show whether it's "bold", "extra-bold", "regular", etc... but it just shows the numeric font-weight. Maybe we should consider allowing a numeric font-weight to be passed in (i.e `fontWeight={400}` instead so that our code more closely matches the designs. I find myself having to look inside this component to see which font weight to use.

HenryQuillin (approved):

escobarjonatan (approved):

PR #78, opened 2022-06-04
Hotfix: updates Typography to use number font-weight values

scandycuz:
- Updates the Typography component to take number values for the `fontWeight` prop, since that matches what UX is using


HenryQuillin (approved):
LGTM

PR #90, opened 2022-07-12
Running format on files and updating spacing on touched files

escobarjonatan:


HenryQuillin:
Let's discuss using theme.spacing vs the default number values here: https://projectnewmgroup.slack.com/archives/C033DLCBGSK/p1658108292107289

HenryQuillin (changes requested):
Let's discuss using theme.spacing vs the default number values here: https://projectnewmgroup.slack.com/archives/C033DLCBGSK/p1658108292107289

src/components/Toast.tsx
```diff
@@ -37,20 +35,24 @@ const Toast = () => {
...
+          success: <CheckCircleIcon />,
         } }
         onClose={ handleClose }
         severity={ severity }
         sx={ {
           "&.MuiAlert-root": {
             ".MuiAlert-message": {
-              paddingRight: ["8px", "16px", "48px"]
+              paddingRight: [
+                theme.spacing(1),
```
HenryQuillin, line 52: I think you forgot to remove the theme.spacing here. Not a big deal tho

src/components/Toast.tsx
```diff
@@ -37,20 +35,24 @@ const Toast = () => {
...
-          success: <CheckCircleIcon />
+          success: <CheckCircleIcon />,
         } }
         onClose={ handleClose }
         severity={ severity }
         sx={ {
           "&.MuiAlert-root": {
             ".MuiAlert-message": {
-              paddingRight: ["8px", "16px", "48px"]
+              paddingRight: [
```
HenryQuillin, line 51: Going forward, I think we should use either p, pr, pl... or padding, paddingRight, paddingLeft.... My vote is for the former for brevity.

HenryQuillin (approved):
LGTM!

src/components/Toast.tsx
escobarjonatan: there are soooooo many ways to do the same thing lol i didnt know you could use the numbers just like that, i actually prefer the numbers as well.... will be changing it soon

src/components/Toast.tsx
escobarjonatan: We can definitely talk about it, these are the cons of us having the flexibility to do it multiple ways :/

PR #95, opened 2022-07-29
Updating toast message when there is an error with social accounts

escobarjonatan:
image

### DIFF
--- a/src/elements/skeletons/TableSkeleton.tsx
+++ b/src/elements/skeletons/TableSkeleton.tsx
@@ -26,7 +26,9 @@ const SkeletonTable: FunctionComponent<SkeletonTableProps> = ({
         display="grid"
         gridTemplateColumns={ `repeat(${cols}, minmax(72px, 1fr))` }
       >
-        { Array(cols).fill(<Skeleton height="40px" width="72px" />) }
+        { Array.from({ length: cols }, (_, i) => (
+          <Skeleton key={ i } height="40px" width="72px" />
+        )) }
       </Stack>
 
       <Stack
@@ -38,7 +40,9 @@ const SkeletonTable: FunctionComponent<SkeletonTableProps> = ({
         justifyContent="space-around"
         gridTemplateColumns={ `repeat(${cols}, minmax(100px, 1fr))` }
       >
-        { Array(cols * rows).fill(<Skeleton height="24px" />) }
+        { Array.from({ length: cols * rows }, (_, i) => (
+          <Skeleton key={ i } height="24px" />
+        )) }
       </Stack>
       <Stack
         display="flex"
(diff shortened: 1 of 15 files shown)


j-mateo:
Task linked: [CU-382cp84 Error toast message for social sign-up account email conflict](https://app.clickup.com/t/382cp84)

src/elements/skeletons/TableSkeleton.tsx
```diff
@@ -38,7 +40,9 @@ const SkeletonTable: FunctionComponent<SkeletonTableProps> = ({
         justifyContent="space-around"
         gridTemplateColumns={ `repeat(${cols}, minmax(100px, 1fr))` }
       >
-        { Array(cols * rows).fill(<Skeleton height="24px" />) }
+        { Array.from({ length: cols * rows }, (_, i) => (
+          <Skeleton key={ i } height="24px" />
+        )) }
```
escobarjonatan, line 45: @HenryQuillin I thought i came up with a cleaner solution to list rendering but forgot the key 🙄
Added it to this PR since it's fairly small and other PR is merged

HenryQuillin: Nice! Yeah I like this better

HenryQuillin (approved):
LGTM!

scandycuz (changes requested):
LGTM, but I think the most important error message to add to add is for when a user tries to sign-up with a social account where the email is already taken. Currently, the back-end will return a 409 error, and the sign-up will fail without diplaying an error message.

I think the best way to do that would be to update the social login API calls with logic to dispatch the `setToastMessage` if it fails. The RTKQuery docs have an [example](https://redux-toolkit.js.org/rtk-query/api/createApi#onquerystarted) of how to implement that. E.g. for the Google login endpoint it could look like:

```
googleLogin: build.mutation<NewmAuthResponse, NewmOAuthRequest>({
  query: (body) => ({
    url: "v1/auth/login/google",
    method: "POST",
    body,
  }),
  async onQueryStarted(id, { dispatch, queryFulfilled }) {
    try {
      const { data } = await queryFulfilled
    } catch (err) {
      const errorMessage = err.status === 409 
        ? "The email for this account is already in use"
        : "An error occurred while logging in";
        
      dispatch(
        setToastMessage({
          message: errorMessage,
          severity: "error",
        })
      )
    }
  },
}),
```

The logic could probably be extracted into a util function so that it can be used for all of the social login endpoints. Also, the existing logic that was handling the social login error should be removed from `src/modules/session/slice.ts`

escobarjonatan:
Abandoning in favor of this new PR: https://github.com/projectNEWM/newm-artist-portal/pull/97

PR #100, opened 2022-08-19
Updating the initial login page with updated mocks

escobarjonatan:
Video Demo: https://user-images.githubusercontent.com/11530120/185542464-08382fd8-6edc-4fe0-8082-6ec00c4bd257.mov

[Figma mocks](https://www.figma.com/file/zrn01ZR2zFeLiKBebEYI1v/Artist-Portal---Final-Designs?node-id=898%3A7448)

Notes: 
- Logo vertical spacing is different than mock so that it can align with the spacing from the rest of the pages in order to fit content in a 13'' macbook.
- "Create new account" button placement and look is different than the mocks in order to follow semantics and consistency within our apps. This also opened up a question about the different appearances of our components. I figured we could make any button related changes once we know what our buttons/links in our app.
- "Forgot password?" button style is different as it is a secondary action to the login button, final decision is still TBD so this is temporary.

### DIFF
--- a/src/components/FacebookLogin.tsx
+++ b/src/components/FacebookLogin.tsx
@@ -34,21 +34,10 @@ const FacebookLogin: FunctionComponent = () => {
     dispatch(sessionApi.endpoints.facebookLogin.initiate({ accessToken }));
   };
 
-  const handleFacebookLoginFail = () => {
-    dispatch(
-      setToastMessage({
-        heading: "Facebook",
-        message: "Facebook authentication was not successful.",
-        severity: "error",
-      })
-    );
-  };
-
   return (
     <FacebookLoginHelper
       appId={ process.env.REACT_APP_FACEBOOK_CLIENT_ID || "" }
       onSuccess={ handleFacebookLoginSuccess }
-      onFail={ handleFacebookLoginFail }
       render={ ({ onClick }) => (
         <IconButton
           aria-label="facebook authorization"
--- a/src/components/FacebookLogin.tsx
+++ b/src/components/FacebookLogin.tsx
@@ -11,6 +11,7 @@ import FacebookLoginHelper, {
 } from "@greatsumini/react-facebook-login";
 import FacebookIcon from "assets/images/FacebookIcon";
 import { IconButton } from "@mui/material";
+import theme from "theme";
 
 const FacebookLogin: FunctionComponent = () => {
   const dispatch = useDispatch();
@@ -21,6 +22,7 @@ const FacebookLogin: FunctionComponent = () => {
     if (!accessToken) {
       dispatch(
         setToastMessage({
+          heading: "Facebook",
           message: "Facebook authentication service offline.",
           severity: "error",
         })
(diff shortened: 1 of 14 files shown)


src/components/FacebookLogin.tsx
```diff
@@ -34,21 +34,10 @@ const FacebookLogin: FunctionComponent = () => {
...
-        severity: "error",
-      })
-    );
-  };
-
   return (
     <FacebookLoginHelper
       appId={ process.env.REACT_APP_FACEBOOK_CLIENT_ID || "" }
       onSuccess={ handleFacebookLoginSuccess }
-      onFail={ handleFacebookLoginFail }
```
escobarjonatan, line 48: Removing this in favor of handling errors on our end through `onQueryStarted` in session api

src/components/GoogleLogin.tsx
```diff
@@ -68,7 +58,6 @@ const GoogleLogin: FunctionComponent = () => {
         </IconButton>
       ) }
       onSuccess={ handleGoogleRespSuccess }
-      onFailure={ handleGoogleRespFailure }
```
escobarjonatan, line 62: Removing this in favor of handling errors on our end through `onQueryStarted` in session api. Google will try to login on initial load and if it doesn't it'll throw this error this way

src/components/LinkedInLogin.tsx
```diff
@@ -22,17 +21,6 @@ const LinkedInLogin: FunctionComponent = () => {
     clientId: process.env.REACT_APP_LINKEDIN_CLIENT_ID || "",
     redirectUri,
     scope: "r_liteprofile r_emailaddress",
-    onError: ({ error }) => {
```
escobarjonatan, line 24: Removing this in favor of handling errors on our end through `onQueryStarted` in session api

src/modules/session/thunks.ts
```diff
@@ -72,21 +72,10 @@ export const handlelLoginError = createAsyncThunk(
...
-      case 409:
-        errorMessage = "The email for this account is already in use";
-        break;
-      default:
-        errorMessage = "An error occurred while logging in";
-    }
+    const errorMessage =
+      error?.status === 409
+        ? "The email for this account is already in use"
+        : "Email or password is incorrect";
```
escobarjonatan, line 78: Removing the more detailed error message and making a generic error message instead

src/pages/login/Login.tsx
```diff
@@ -35,8 +35,8 @@ const Login: FunctionComponent = () => {
     setMaskPassword(!maskPassword);
   };
 
-  const handleLogin = ({ email, password }: FormikValues) => {
-    dispatch(sessionApi.endpoints.login.initiate({ email, password }));
+  const handleLogin = async ({ email, password }: FormikValues) => {
+    await dispatch(sessionApi.endpoints.login.initiate({ email, password }));
```
escobarjonatan, line 39: Made this async/await so that Formik knows when to turn `isSubmitting` to true/false and disable the login button to prevent it from being executed multiple times while the call is being made.

escobarjonatan:
> Looks great. I was noticing that only the text for the "Forgot password" button is clickable though. It looks like it's an existing issue with the `OutlinedButton` component, but it would be good to update the entire button to be clickable.

@scandycuz  I noticed that and the reason why i didn't fix it was because due to the slack conversations we are having about buttons. I see some refactor coming to it soon and i was just going to fix it there. I can do it if you think it's worth it though

PR #101, opened 2022-08-29
Refactor RTKQuery success / error handling

scandycuz:
Refactors RTK query requests so that success/error logic is handled using `onQueryStarted`.


cullenpu (approved):
In line with what we discussed last week👍

PR #106, opened 2022-09-15
[DO NOT MERGE - ONLY FOR CODE REVIEW] MURS token drop mini-site

scandycuz:
Branch for MURS token drop mini-site.

Updates to the MURS token drop app should be merged into this branch.

<table>
<tr>
<td>Screen Shot 2022-09-14 at 6 46 38 PM</td>
<td>Screen Shot 2022-09-14 at 6 49 43 PM</td>
<td>Screen Shot 2022-09-14 at 6 48 55 PM</td>
</tr>
</table>

### DIFF
--- a/src/common/stringUtils.ts
+++ b/src/common/stringUtils.ts
@@ -1,3 +1,0 @@
-export const capitalize = (value: string) => {
-  return value[0].toUpperCase() + value.slice(1);
-};
--- a/src/common/index.ts
+++ b/src/common/index.ts
@@ -7,5 +7,4 @@ export * from "./urlUtils";
 export * from "./fileUtils";
 export * from "./types";
 export * from "./urlUtils";
-export * from "./stringUtils";
 export * from "./mockConstants";


src/elements/GradientCircularProgress.tsx
```diff
@@ -0,0 +1,42 @@
...
+        <linearGradient id="linearColors" x1="0" y1="0" x2="1" y2="1">
+          <stop offset="20%" stopColor={ theme.colors[startColor] } />
+          <stop offset="90%" stopColor={ theme.colors[endColor] } />
+        </linearGradient>
+      </svg>
+
+      <CircularProgress
+        thickness={ 4 }
+        classes={ { circle: classes.circle } }
+        // { ...props }
```
escobarjonatan, line 36: Was this left commented out on purpose?

src/elements/Link.tsx
```diff
@@ -5,20 +5,27 @@ import {
...
+      replace={ replace }
+      style={ { textDecoration: underline ? "underline" : "none" } }
+    >
       <Typography
         { ...typographyProps }
         ref={ ref }
         color={ color }
         sx={ {
-          textDecoration: "underline",
+          textDecoration: underline ? "underline" : "none",
```
escobarjonatan, line 28: Not sure if you were trying to make the underline an explicit prop, but if you weren't you can pass it into the sx prop and it will override the default like so:

`<Link other="props" sx={ { textDecoration: "none" } }>some text</Link>`

src/pages/signUp/Welcome.tsx
```diff
@@ -1,7 +1,6 @@
 import { Box, Stack, useTheme } from "@mui/material";
 import { FilledButton, Typography } from "elements";
 import { FunctionComponent, useState } from "react";
```
escobarjonatan, line 3: If you are removing this page from the route, i am wondering if this was necessary? or just making sure we dont authenticate for sure

src/pages/tokenDrop/Congratulations.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { FunctionComponent } from "react";
+import { useNavigate } from "react-router-dom";
+import Confetti from "react-confetti";
+
+const Congratulations: FunctionComponent = () => {
+  const theme = useTheme();
+  const navigate = useNavigate();
+  const window = useWindowDimensions();
+  const isMdScreen = useMediaQuery(
+    `(min-width:${theme.breakpoints.values.md}px)`
```
escobarjonatan, line 17: what does the "min-width:" do in this case? I have used the following method to try to figure out the breakpoint and trying to figure out if this is just another way to do the same thing:

`const isBelowMediumBreakpoint = useMediaQuery(theme.breakpoints.down("md"));`

src/pages/tokenDrop/Footer.tsx
```diff
@@ -0,0 +1,86 @@
...
+      to: "#",
+      label: "Disclaimer",
+      type: "router",
+    },
+    {
+      to: "#",
+      label: "Terms of Service",
+      type: "router",
+    },
+  ];
```
escobarjonatan, line 40: Nice approach 👏🏽

src/pages/tokenDrop/Landing.tsx
```diff
@@ -0,0 +1,203 @@
...
+                  >
+                    <PlayIcon />
+                  </IconButton>
+                ) }
+              </Box>
+
+              <img
+                src={ mursProfileImageXs }
+                style={ { width: 40, height: 40, borderRadius: "50%" } }
+                alt="murs profile"
```
escobarjonatan, line 130: MURS* adding a comment since this is not text that is shown automatically

src/pages/tokenDrop/Landing.tsx
```diff
@@ -0,0 +1,203 @@
...
+                  Break up
+                </Typography>
+                <Typography variant="subtitle2">MURS</Typography>
+              </Box>
+            </Stack>
+
+            <a
+              href="https://open.spotify.com/track/1ZSwTbIdB0p7pNLDck5RQb?si=b1c674b4020d4e90"
+              target="_blank"
+              rel="noreferrer"
```
escobarjonatan, line 144: **Action Requested:**
I believe starting 2021 modern browser versions now protect against a major vulnerability and add noopener by default, but i am wondering if we should add that to our links to protect users in older browsers? so it would be: `rel="noopener noreferrer"`

src/common/stringUtils.ts
```diff
@@ -0,0 +1,3 @@
+export const capitalize = (value: string) => {
```
escobarjonatan, line 1: Do we need this? I think that this can be done with [css text-transform](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform)

PR #108, opened 2022-09-18
Improves purchase flows

scandycuz:
- Disables "Purchase" button only after transaction is submitted (instead of after purchase order is created)
- Updates UI messaging around creating a transaction using a wallet extension


escobarjonatan (approved):

src/pages/tokenDrop/Payment.tsx
```diff
@@ -42,12 +44,23 @@ const Payment: FunctionComponent = () => {
   const dispatch = useDispatch();
 
   const bundlePrice = useGetMursPrice();
```
AndrewWestberg, line 45: We should remove references to "murs" in case we re-use this code for other artists.

AndrewWestberg (approved):

PR #109, opened 2022-09-18
Adding modal and content for artist socials and faq section

escobarjonatan:
https://user-images.githubusercontent.com/11530120/190932803-de6b12b9-7f4d-4649-8a9f-f5b14ef7e105.mov


src/pages/tokenDrop/Footer.tsx
```diff
@@ -1,41 +1,70 @@
...
 
 const Footer: FunctionComponent = () => {
   const theme = useTheme();
+  const [isOpen, setIsOpen] = useState(false);
+  const [modalId, setModalId] = useState("");
+
+  const handleClose = () => setIsOpen(false);
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const handleOpen = (event: any) => {
+    setModalId(event?.target?.getAttribute("id"));
```
escobarjonatan, line 23: So the reason why i am doing this is to get the id attribute of the link that was clicked so that i can set the modal content depending on that. I feel like there's a better way to do it by using `useRef` but it was giving me an error about forwarding ref, this is something we can look at in the future

PR #113, opened 2022-09-20
Modify modal background and updated artists links

escobarjonatan:


src/pages/tokenDrop/SocialsModal.tsx
```diff
@@ -34,13 +34,36 @@ const SocialsModal = () => (
           flexWrap: "wrap",
         } }
       >
-        <AccentButton startIcon={ <TwitterIcon /> }>Twitter</AccentButton>
-        <AccentButton startIcon={ <InstagramIcon /> }>Instagram</AccentButton>
-        <AccentButton
-          startIcon={ <SpotifyIcon height={ 16 } width={ 16 } fillColor="#DC3CAA" /> }
+        <a
```
escobarjonatan, line 37: The reason behind wrapping the button in a link is that we needed to add the `rel` and `target` properties and i didnt see that MUI button component supported that atm so this is a workaround for that. Something to keep in mind as we refactor the buttons in the future

PR #120, opened 2022-09-26
Updating modal copy and adding subscribe functionality to be reused in congrats page

escobarjonatan:
Updates copy for modal

Need to figure out what we want the subscribe functionality to do, will update once i know


escobarjonatan:
@scandycuz Requested changes have been pushed

PR #125, opened 2022-09-30
MURS App: countdown

scandycuz:
- Adds countdown page
- Updates content with MURS bio

### DIFF
--- a/src/buildParams.ts
+++ b/src/buildParams.ts
@@ -5,7 +5,7 @@ export const isMainNet = process.env.REACT_APP_NETWORK_MODE === "1";
 export const enableReduxLogging = true;
 
 // set this to false to disable countdown page
-export const enableCountdown = true;
+export const enableCountdown = false;
 
 // TODO: Update with NEWM server production url host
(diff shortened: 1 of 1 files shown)


src/buildParams.ts
```diff
@@ -19,21 +22,25 @@ export const baseUrls: Record<string, string> = {
 // Update values here for the artist being featured
 export const projectDetails = {
   artistName: "MURS",
+  launchDate: "October 8, 2022 00:00:00",
```
escobarjonatan, line 25: Would it be easier to get the data from the api in case it changes and we dont have to change it every release?

```
{
  "status": "error",
  "message": "Sale is closed!",
  "data": [
    "com.firehose.model.ProjectResponse",
    {
      "cost": 2000000,
      "tokenCostType": null,
      "tokenCost": null,
      "startDate": 1663373418000,
      "endDate": null,
      "dripCampaigns": null
    }
  ]
}

```

escobarjonatan (approved):

PR #149, opened 2022-10-09
Token sale app: removes Nami as supported wallet

scandycuz:


src/modules/wallet/utils.ts
```diff
@@ -24,13 +24,7 @@ import { CreateTransactionParams, EnabledWallet, Wallets } from "./types";
 // 1 = Mainnet, 0 = Testnet
 const networkMode = isProd ? 1 : 0;
 
-export const supportedWallets = [
```
escobarjonatan, line 27: Not sure if this is temporary or not but if it's permanent we should remove all assets related to nami

AndrewWestberg: Hopefully, it's temporary. Nami is a single-address wallet. We currently had some code calling getUnusedAddress(). Nami would return an empty array since there's only a single address(used) in it. We probably just need to modify the code in the future so we don't call getUnusedAddress() anymore.

escobarjonatan (approved):

PR #151, opened 2022-10-13
Token sale - select bundle size

scandycuz:
- Adds back-up wallet address selection for Nami
- Implements a drop-down to select multiple bundles

### DIFF
--- a/src/modules/sale/utils.ts
+++ b/src/modules/sale/utils.ts
@@ -15,3 +15,9 @@ export const parseBundleAmounts = (
 
   return { size, usdPrice, adaPrice, royaltyPercentage };
 };
+
+export const parsePurchasePrice = (value?: string) => {
+  if (!value) return "UNDEFINED";
+
+  return value.slice(1).split(".")[0];
+};
--- a/src/pages/tokenDrop/Landing.tsx
+++ b/src/pages/tokenDrop/Landing.tsx
@@ -182,6 +182,7 @@ const Landing: FunctionComponent = () => {
                   label=""
                   name="bundles"
                   value={ selectedBundleId }
+                  placeholder="1 bundle"
                   handleChange={ (option) => {
                     dispatch(setSelectedBundleId(option.value));
                   } }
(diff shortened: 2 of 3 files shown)


src/buildParams.ts
```diff
@@ -51,8 +50,7 @@ const projectDetailsMap: Record<string, ProjectDetails> = {
     launchTimestamp: 1665212400000,
     projectId: Number(process.env.REACT_APP_PROJECT_ID) || 6,
     bundleAmount: 8000,
-    totalBundles: 10000,
-    bundlePercentage: 0.8,
+    totalTokens: 100000000,
```
AndrewWestberg, line 53: If you're removing the bundlePercentage, shouldn't totalTokens be 80,000,000?

AndrewWestberg: Ignore this. Looks like this is used to calculate the royaltyPercentage. In which case, 100m is correct.

src/modules/wallet/utils.ts
```diff
@@ -201,11 +201,23 @@ export const createTransaction = async (body: CreateTransactionParams) => {
...
-  const addresses = await window.Wallets[walletName].getUnusedAddresses();
-  const encodedAddress = addressFromHex(addresses[0]);
+export const getWalletAddress = async (walletName: string): Promise<string> => {
+  let addresses = [];
+  const wallet = window.Wallets[walletName];
+
+  // prefer unused address, but some wallets such as Nami require used address
+  addresses = await wallet.getUnusedAddresses();
+  if (addresses.length === 0) {
+    addresses = await wallet.getUsedAddresses();
```
AndrewWestberg, line 211: Up to you, but this could be simplified to just always call `getUsedAddresses()` regardless of whether its Nami or not. When we end up sending tokens back to them, we will always be sending to a used address anyway.

Sending to the very first address in the wallet prevents a specific type of attack called a franken-address attack. Our current sale doesn't make sense to attack this way, but if we ever implement something like whitelisting, it prevents someone from hacking themselves onto the whitelist pretending to be someone else.

AndrewWestberg (approved):

PR #154, opened 2022-10-15
Updating intro copy, adding a tooltip and an agreement link

escobarjonatan:


src/pages/tokenDrop/Landing.tsx
```diff
@@ -81,6 +81,19 @@ const Landing: FunctionComponent = () => {
         <Typography variant="subtitle1" sx={ { whiteSpace: "pre-wrap" } }>
           { projectDetails.description }
         </Typography>
+        <Typography variant="subtitle1" sx={ { mt: 2, whiteSpace: "pre-wrap" } }>
+          Explore the legal side of how Stream Tokens and royalties work in this
```
escobarjonatan, line 85: @scandycuz The only thing that was different between the sales was the link where the "here" points to. I wanted to stay away from adding a third party library for now. If we need to change this in the future to use a string to html parser we can.

src/pages/tokenDrop/Landing.tsx
```diff
@@ -92,84 +105,96 @@ const Landing: FunctionComponent = () => {
...
-              <Box>
-                <Typography variant="h4" fontWeight={ 700 }>
-                  { projectDetails.songName }
-                </Typography>
-                <Typography variant="subtitle2">
-                  { projectDetails.artistName }
-                </Typography>
-              </Box>
-            </Stack>
+          <Box display="flex" alignItems="center">
```
escobarjonatan, line 108: I added this box as a wrapper to be able to add a new tooltip in this section idk why git thinks i modified a lot here

PR #158, opened 2022-10-25
Updating redirect when the sale has ended

escobarjonatan:


src/modules/sale/api.ts
```diff
@@ -46,9 +49,16 @@ const extendedApi = api.injectEndpoints({
...
+          // redirect to sold out page if the sale has ended
+          if (isSaleEnded) {
+            if (!isSoldoutCurrentPage) window.location.pathname = "sold-out";
+            return;
+          }
+
           // sale is sold out, redirect to sold out page
-          if (isSoldout && !isSoldoutCurrentPage) {
-            window.location.pathname = "sold-out";
+          if (isSoldout) {
```
escobarjonatan, line 60: `isSaleEnded` and `isSoldout` could be combined but i decided not to since `isSaleEnded` will change in the short future and it will be its own conditional

PR #159, opened 2022-10-29
Creating new button component with new brand colors and styles

escobarjonatan:
Created a new button component with new brand colors and styles.

Note: 
This PR is just to handle the creation of the component, a follow-up PR will be created to delete/replace other button components throughout our artist portal  

[Figma link](https://www.figma.com/file/o0dRbDaCTNOKZDx5j5PmMu/Design-System?node-id=1%3A2)

I've added a couple of stories for this button so that we can play around with the configuration that is available to our component

Preview:
https://user-images.githubusercontent.com/11530120/198852685-9f5cae18-13d0-448c-990f-c69f6ddefb9f.mov

### DIFF
--- a/src/elements/Button.tsx
+++ b/src/elements/Button.tsx
@@ -118,27 +118,3 @@ const Button: FunctionComponent<ButtonProps> = ({
 };
 
 export default Button;
-
-/*
-- DONE - Width
-    - default (determined by content inside)
-    - Full width
-    - N/A - Max width 340px (Should be handled by implementer, they can pass in sx max-width: 340px)
-- DONE - Icon. Will be handled with MUI button as it supports it out of the box
-    - Before content
-    - After content
-    - In the center (button icon like, loading state for example)
-- DONE - Anchor type. Will be handled with MUI button as it supports it out of the box
-    - href and other properties checkout mdn
-- DONE - Variant
-    - Primary (default)
-    - Secondary
-    - Outlined
-- DONE - Color
-    - Copy from theme
-    - Potentially add config to add a gradient button ?
-
-TODO:
-- Delete transparent button
-- push changes to this pr and then create a new PR
-*/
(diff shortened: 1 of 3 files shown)


src/elements/Button.tsx
```diff
@@ -0,0 +1,144 @@
...
+      minWidth: "unset",
+      p: 0,
+      width: "40px",
+    },
+  };
+
+  const variantStyles = {
+    primary: {
+      background:
+        color === "white" ? theme.gradients.music : theme.gradients[color],
```
escobarjonatan, line 58: Since we dont have a "white" linear gradient for our primary buttons, i set it to default to our music gradient

src/elements/Button.tsx
```diff
@@ -0,0 +1,144 @@
...
+        textTransform: "none",
+        "&.Mui-disabled": {
+          color: theme.colors[color],
+        },
+        ...widthStyles[width],
+        ...variantStyles[variant],
+      } }
+      { ...rest }
+    >
+      { isLoading ? (
```
escobarjonatan, line 104: @cullenpu  I've implemented the logic you did for the `filledButton` here since it will be replaced with this component soon. Other changes in your PR i feel still apply since that's dealing with the app state

src/elements/Button.tsx
```diff
@@ -0,0 +1,120 @@
+import { FunctionComponent } from "react";
+import {
+  CircularProgress,
+  Button as MUIButton,
+  ButtonProps as MUIButtonProps,
+} from "@mui/material";
+import theme from "theme";
+
+interface ButtonProps extends Omit<MUIButtonProps, "color" | "variant"> {
+  readonly color?:
```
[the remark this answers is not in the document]
escobarjonatan: I thought about doing that, the reason i didnt was because the mocks didnt have a "newm" button variant but had a "white" (which is not in the gradients). Currently, every gradient color specified here has a solid color equivalent , with the exception of "white" but added some logic to point to the music gradient. Going to leave it as is for now and we can update it later if we need to

PR #161, opened 2022-10-31
Token sale app - expandable landing description

scandycuz:
Abbreviates the landing page description with an option to "Read more" and view the full content.

### DIFF
--- a/src/components/ReadMore.tsx
+++ b/src/components/ReadMore.tsx
@@ -7,6 +7,9 @@ interface ReadMoreProps {
   readonly expandedContent: ReactNode;
 }
 
+/**
+ * Displays abbreviated content with an option to expand and "read more".
+ */
 const ReadMore: FunctionComponent<ReadMoreProps> = ({
   abbreviatedContent,
   expandedContent,
--- a/src/pages/tokenDrop/Landing.tsx
+++ b/src/pages/tokenDrop/Landing.tsx
@@ -14,7 +14,7 @@ import PlayIcon from "assets/images/PlayIcon";
 import SpotifyIcon from "assets/images/SpotifyIcon";
 import { useNavigate } from "react-router-dom";
 import StopIcon from "assets/images/StopIcon";
-import { DisplayText, SectionHeading } from "components";
+import { DisplayText, ReadMore, SectionHeading } from "components";
 import { Howl } from "howler";
 import { projectDetails } from "buildParams";
 import poolPmIcon from "assets/images/pool-pm-icon.png";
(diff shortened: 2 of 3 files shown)


src/components/ReadMore.tsx
```diff
@@ -0,0 +1,55 @@
+import { Box, useTheme } from "@mui/material";
+import { Typography } from "elements";
+import { FunctionComponent, ReactNode, useState } from "react";
+
+interface ReadMoreProps {
+  readonly abbreviatedContent: ReactNode;
+  readonly expandedContent: ReactNode;
```
escobarjonatan, line 7: What do you think about making the ReadMore component take in 3 props, 
`characterCount: number` = Determine how many characters to show before the readmore logic displays
`text: string` = the full content to display
`typographyOverrides: object` = overwrite the props for the MUI typography component this would ultimately render 

The benefit of doing it this way would be so we can then do the manipulation/logic inside of this component and keep the parent components a little cleaner

src/pages/tokenDrop/Landing.tsx
```diff
@@ -84,22 +84,33 @@ const Landing: FunctionComponent = () => {
...
+          abbreviatedContent={
+            <Typography variant="subtitle1" sx={ { whiteSpace: "pre-wrap" } }>
+              { projectDetails.description
+                .split("\n\n")
+                .slice(0, 2)
+                .join("\n\n") }
+            </Typography>
+          }
+          expandedContent={
+            <Typography variant="subtitle1" sx={ { whiteSpace: "pre-wrap" } }>
```
escobarjonatan, line 97: I would think you always want to display this content as this is not part of the description and more of a legal information we wouldn't want to hide from our users

src/components/ReadMore.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Displays abbreviated content with an option to expand and "read more".
+ */
+const ReadMore: FunctionComponent<ReadMoreProps> = ({
+  abbreviatedContent,
+  expandedContent,
+}) => {
+  const theme = useTheme();
```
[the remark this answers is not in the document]
escobarjonatan: I would think importing from a static file vs a processed place like a hook we get some sort of performance gain

src/components/ReadMore.tsx
escobarjonatan: I dont think we need to handle the 2 different sales individually though... the rest of the content is the same, if anything i think it would help to make them display the same amount of characters(round down to the word) for consistency since it would take similar amount of space.

escobarjonatan: Yea i dont see it being something different than a typography component that would need a read more logic

escobarjonatan: I think there's a lot of room for interpretation in our mocks as we've seen previously though

escobarjonatan (approved):

PR #172, opened 2022-12-05
Adding initial stricture to edit song page and library

escobarjonatan:
[Figma link](https://www.figma.com/file/zrn01ZR2zFeLiKBebEYI1v/Artist-Portal---Final-Designs?node-id=1023%3A7926&t=pYwSX55dnO1z6iUl-0)

https://user-images.githubusercontent.com/11530120/205725044-278a32ae-e830-422d-8e50-07ff34f9ce68.mov


src/assets/images/VerticalEllipsis.tsx
```diff
@@ -1,17 +1,14 @@
-import * as React from "react";
-
 const VerticalEllipsis = (props: any) => (
```
escobarjonatan, line 1: Updated this SVG so that it does not have predefined styles

src/components/home/SideBarNavLink.tsx
```diff
@@ -38,7 +38,7 @@ const SideBarNavLink: FunctionComponent<SideBarNavLinkProps> = ({
   onClick,
 }) => {
   const resolved = useResolvedPath(to || "");
-  const match = useMatch(resolved.pathname);
```
escobarjonatan, line 41: Updated this match so that any url path with example "/library/*" makes the "library" section of the sidenav as selected

src/components/styled/StyledPagination.tsx
```diff
@@ -19,7 +19,7 @@ const StyledPagination = styled(Pagination)<PaginationProps>({
   "& .MuiPaginationItem-previousNext": {
     borderRadius: "0px 6px 6px 0px",
   },
-  "& li:first-child > button": {
+  "& li:first-of-type > button": {
```
escobarjonatan, line 22: updated this selector so that it is SSR friendly as `first-child` can change between hydration

src/pages/home/library/Discography.tsx
```diff
@@ -0,0 +1,90 @@
+import { Box } from "@mui/material";
```
escobarjonatan, line 1: The content in here should be similar to what was in `Library.tsx`, needed to extract this to handle routing in new `Library.tsx` file

src/pages/home/library/EditSong.tsx
```diff
@@ -0,0 +1,120 @@
...
+  children: ReactNode;
+  index: number;
+  value: number;
+}
+
+interface ColorMap {
+  [index: number]: Partial<keyof Theme["gradients" | "colors"]>;
+}
+
+const TabPanel = ({ children, value, index }: TabPanelProps) => {
```
escobarjonatan, line 21: I could extract out the Tabs/TabPanel into its own component if you all feel that we will have a good use in the future?

src/pages/home/library/SongList.tsx
```diff
@@ -144,22 +140,22 @@ export default function SongList({
...
                   </StyledTableCell>
                   <StyledTableCell align="right">
                     <Box sx={ { display: { xs: "none", sm: "inline" } } }>
-                      { row.createdAt.slice(0, 10) }{ " " }
+                      { song.createdAt.slice(0, 10) }{ " " }
                     </Box>
-                    <IconButton>
-                      <VerticalEllipsis />
-                    </IconButton>
+                    <Link to="edit-song" state={ { ...song } }>
```
escobarjonatan, line 156: Replaced it to this since we aren't sure how we want to handle the navigation to the song yet. Once that is figured out we can delete this and apply a desired solution

src/assets/images/VerticalEllipsis.tsx
```diff
@@ -1,17 +1,14 @@
...
     fill="none"
     xmlns="http://www.w3.org/2000/svg"
-    { ...props }
   >
-    <rect width={ 36 } height={ 36 } rx={ 18 } fill="#1C1C1E" />
     <path
-      d="M18 10.5c-.917 0-1.667.75-1.667 1.667 0 .916.75 1.666 1.667 1.666s1.667-.75 1.667-1.666c0-.917-.75-1.667-1.667-1.667Zm0 11.667c-.917 0-1.667.75-1.667 1.666 0 .917.75 1.667 1.667 1.667s1.667-.75 1.667-1.667c0-.916-.75-1.666-1.667-1.666Zm0-5.834c-.917 0-1.667.75-1.667 1.667s.75 1.667 1.667 1.667 1.667-.75 1.667-1.667-.75-1.667-1.667-1.667Z"
-      fill="#fff"
+      d="M1.99984 0.375C1.129 0.375 0.416504 1.0875 0.416504 1.95833C0.416504 2.82917 1.129 3.54167 1.99984 3.54167C2.87067 3.54167 3.58317 2.82917 3.58317 1.95833C3.58317 1.0875 2.87067 0.375 1.99984 0.375ZM1.99984 11.4583C1.129 11.4583 0.416504 12.1708 0.416504 13.0417C0.416504 13.9125 1.129 14.625 1.99984 14.625C2.87067 14.625 3.58317 13.9125 3.58317 13.0417C3.58317 12.1708 2.87067 11.4583 1.99984 11.4583ZM1.99984 5.91667C1.129 5.91667 0.416504 6.62917 0.416504 7.5C0.416504 8.37083 1.129 9.08333 1.99984 9.08333C2.87067 9.08333 3.58317 8.37083 3.58317 7.5C3.58317 6.62917 2.87067 5.91667 1.99984 5.91667Z"
+      fill="white"
```
dmkirshon, line 11: I'm adding a few new icons for the Minting and Marketplace columns, with that I will make sure that the styles aren't predefined. 

For the fill for now though I'm importing `theme.tsx` and using `theme.colors`. But I noticed when working with the Sidebar Icons that the fill was not customizable, unless I wasn't configuring the correct css? I had to port the previously used transparent button opacity css fix vs just making the SVG fill of the unhighlighted NavLink `grey 100`

During the research session with Dr. Google, I did find this fix for React SVGs where you pass in something like `currentColor` as a prop for the fill, and default `theme.colors.white`. Let me know if this sounds worth applying to the current Icons we are working on, and then creating a new task to fix the remaining ones and adding an Icon SVG type/interface.

dmkirshon (approved):
No issues here, besides one suggestion for SVG Icons.

src/assets/images/VerticalEllipsis.tsx
escobarjonatan: Good catch, when adding SVGs we should try to avoid adding a fill color unless we have absolutely have to. Example of a have to imo: Spotify icon, you will most likely want to keep it the same brand color throughout.

For this, I will be adding a `fill:"currentColor"` as you mentioned, and that will allow us to set the text color through the component that is using this icon. In this case, the `color` css property is already white from the button component used in `EditSong`. 

I think we should follow this pattern going forward with adding new SVGs.

PR #176, opened 2022-12-19
Edit song info tab, adding ability to modify uploaded song information

escobarjonatan:
This is an initial implementation to modifying song information.

[Figma link
](https://www.figma.com/file/zrn01ZR2zFeLiKBebEYI1v/Artist-Portal---Final-Designs?node-id=1023%3A8463&t=eJnHs6eu6MJzUzQl-1)

https://user-images.githubusercontent.com/11530120/208541440-32de3c26-ef67-49ec-b959-21b573d4f5de.mov


src/pages/home/library/EditSong.tsx
```diff
@@ -1,12 +1,13 @@
-import { Box, Stack, Tab, Tabs, Theme, Typography } from "@mui/material";
+import { Stack, Tab, Tabs, Theme, Typography } from "@mui/material";
+import ArrowBackIcon from "@mui/icons-material/ArrowBack";
+import DeleteIcon from "@mui/icons-material/Delete";
```
escobarjonatan, line 3: Using icons from MUI vs uploading them manually.

I think that for the foreseeable future we should use icons from MUI if possible, because there are a lot of changes happening on the design end that makes it difficult to maintain our internal icons. Once we have a more stable process we can upload our own icons and maintain that. I dont think spending the extra effort is worth it at the moment.

src/buildParams.ts
```diff
@@ -7,6 +7,6 @@ export const enableReduxLogging = !isProd && true;
 export const baseUrls: Record<string, string> = {
   newm: isProd
     ? "https://staging-newm-server.herokuapp.com/"
-    : "https://staging-newm-server.herokuapp.com/",
+    : "https://garage.newm.io/",
```
dmkirshon, line 10: I made this change and the production Url change on my branch as part of the following task, https://app.clickup.com/t/3vc8t71. Not sure how this should be handled, possibly comment/link it to your commit/branch on the ticket and then my PR will add the production Url to finalize?

It's also a breaking change since developers will need to create new accounts? Not sure if that matters much with the stage of development we are in.

escobarjonatan: I think we can both make the change, I will change the prod url to the studio one on my end. If there's a merge conflict we can take care of it when we get to it. 

As far as it being a breaking change, we should be okay to make the change without worrying about that, we add a comment on the `web-dev` channel that we have made this change and we should be good to go.

src/pages/home/library/SongInfo.tsx
```diff
@@ -0,0 +1,226 @@
...
+  const windowWidth = useWindowDimensions()?.width;
+  const { genres } = useSelector(selectContent);
+  const { id = "" } = location.state as Song;
+  const { data = {} } = useGetSongQuery(id);
+
+  const [coverArtUrl, setCoverArtUrl] = useState("");
+  const [description, setDescription] = useState("");
+  const [genre, setGenre] = useState("");
+  const [title, setTitle] = useState("");
+
```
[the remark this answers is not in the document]
escobarjonatan: Well this is awkward haha good catch, for some reason it wasn't working for me without it, but i think it had to do with the reinitialize prop on Formik. I will remove the state and useeffect.

PR #182, opened 2023-02-03
Implementing initial Mint UI in the Library

escobarjonatan:
[Figma link](https://www.figma.com/file/zrn01ZR2zFeLiKBebEYI1v/Artist-Portal---Final-Designs?node-id=2335%3A26655&t=qBH2BxEl72tJkE02-1)

https://user-images.githubusercontent.com/11530120/216526170-22464cf6-fbe3-4a1a-b8be-cdefea749e77.mov


src/pages/home/uploadSong/AddOwnerModal.tsx
```diff
@@ -1,17 +1,49 @@
+import * as Yup from "yup";
```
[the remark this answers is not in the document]
dmkirshon: Woot reuse! Should this be pulled out to a common folder, not sure where modals fit in, components or into a common folder under pages? 
I know there was a large discussion on slack about design flow, so I won't belabor that too much here. However. this would then include all the minting tsx related files under the `pages>uploadSong` file path?

The other piece connected to the design flow that I don't think was mentioned in the discussion on slack was the fact that on `UPLOAD A SONG` when the `MINT SONG` selection is selected, the user needs to connect their wallet. This is in the `MintSongModal.tsx` file, is this going to be added in as well if the `Mint.tsx` is kept separate for edit song?

src/pages/home/uploadSong/AddOwnerModal.tsx
```diff
@@ -1,17 +1,49 @@
...
+      .required()
+      .test(
+        "at-least-one-true",
+        "Rights or Credits must be true",
+        (item, testContext) => item || testContext.parent.isRightsOwner
+      ),
+    isRightsOwner: Yup.boolean(),
+    lastName: commonYupValidation.lastName,
+    role: commonYupValidation.role(roles),
+  });
```
dmkirshon, line 46: I also agree with part of the slack discussions, the `Add` button should be disabled if one of the options requires marking to move on. My opinion, and not a code change requirement at this stage but should confer with design, is that the text should probably still be shown when hovering over the `Add` button or in some other type of info tip since design didn't agree with the included one. Might not be an issue though as design said in the users point of view and have to see how the user input goes.

src/pages/home/library/Mint.tsx
```diff
@@ -0,0 +1,451 @@
...
+                          </Typography>
+                        </Stack>
+                        <Stack flexDirection="row" alignItems="center">
+                          <TextInput
+                            max={ 100 }
+                            min={ 0 }
+                            onChange={ (event) =>
+                              handlePercentageChange(event, owner.email)
+                            }
+                            placeholder="0 %"
```
dmkirshon, line 179: Is there a way to keep the percent symbol visible as the numbers change? Otherwise the `AddOwner.tsx` uses the following within their `TextInputField`:
```
endAdornment={
                <InputAdornment
                  position="start"
                  sx={ {
                    color: theme.colors.white,
                    mr: 1,
                  } }
                >
                  <Typography>%</Typography>
                </InputAdornment>
              }
```

src/pages/home/library/Mint.tsx
```diff
@@ -0,0 +1,451 @@
...
+                } }
+                width={
+                  windowWidth && windowWidth > theme.breakpoints.values.md
+                    ? "compact"
+                    : "default"
+                }
+              >
+                Next
+              </Button>
+            ) }
```
dmkirshon, line 302: Should this button also be disabled if Credits is missing, is it proper to mint a song without a Credits to show on `song detail`? Or pop up an alert, info, or field requirement warning? Also is song detail the `song description ` on the `INFO` tab? Or is song detail just a minting field?

src/pages/home/uploadSong/AddOwnerModal.tsx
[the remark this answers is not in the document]
escobarjonatan: I thought about moving it to the components, the reason why i didnt was because i felt more confident that it would change vs not in the near future. I can add it to the components if you all feel that this will not change soon.

src/pages/home/uploadSong/AddOwnerModal.tsx
escobarjonatan: Yea i went with the error message because that's what we had done for the sale page. seems like design wants something different now(which is my preferable solution). I will be making the change to delete the error text and disable the button

src/pages/home/library/Mint.tsx
escobarjonatan: oh that's an interesting approach, i will check it out and make it consistent.

src/pages/home/uploadSong/AddOwnerModal.tsx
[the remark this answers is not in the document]
escobarjonatan: It should be a pretty easy change to do, for me it was a matter of guessing what was more likely to happen.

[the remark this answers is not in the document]
escobarjonatan: oh wait, i had not seen your other PR, i will not be making that change as part of this PR as i can see how merge conflicts will arise

src/pages/home/library/Mint.tsx
escobarjonatan: >is it proper to mint a song without a Credits to show on song detail?
That right there is a great question, i will ask product and update this accordingly.
>is song detail the song description  on the INFO tab? Or is song detail just a minting field?
This is something that @scandycuz might have more info on. I would assume the song detail and credit song details are 2 separate things? idk though

src/pages/home/library/Mint.tsx
```diff
@@ -0,0 +1,451 @@
...
+          </Stack>
+        </>
+      ) : (
```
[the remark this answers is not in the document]
escobarjonatan: Sure thing we can incorporate them as you see fit

src/pages/home/uploadSong/AddOwnerModal.tsx
escobarjonatan: Disabling the submit button until all required fields are completed is a better option imo, compared to enabling the submit button and showing error messages when the user attempts to submit. Disabling the submit button provides a clear indication of required fields and prevents incomplete submissions, helping to avoid frustration and confusion. On the other hand, enabling the submit button and showing error messages only when clicked can result in confusion and missed error messages, leading to incomplete submissions

src/pages/home/library/Mint.tsx
```diff
@@ -0,0 +1,451 @@
...
+  const [hasAgreedToTerms, setHasAgreedToTerms] = useState(false);
+  const [showWarning, setShowWarning] = useState(true);
+  const [isMintSwitchOn, setIsMintSwitchOn] = useState(false);
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [isFirstStep, setIsFirstStep] = useState(true);
+  const [isFullyOwned, setIsFullyOwned] = useState(false);
+  const [futureOwners, setFutureOwners] = useState<FutureOwner[]>([]);
+  const [futureCreditors, setFutureCreditors] = useState<
+    Partial<FutureOwner>[]
+  >([]);
```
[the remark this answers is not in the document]
escobarjonatan: The reason why i didnt add a formik around this was because i wasn't sure what all was going to be submitted as part of the owners/creditors. I figured we could add it later once we know what the structure would look like?

src/pages/home/library/Mint.tsx
```diff
@@ -0,0 +1,451 @@
...
+
+interface FutureOwner {
+  email: string;
+  firstName: string;
+  isCreator: boolean;
+  isRightsOwner: boolean;
+  lastName: string;
+  percentage: number;
+  role: string;
+}
```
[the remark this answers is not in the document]
escobarjonatan: Good call

src/pages/home/library/Mint.tsx
dmkirshon: I guess looking off of that Danketsu metadata I'm still not sure how everything lines up, might just be a point of education for me on the user interface to minting metadata side, or might be part of an existing MVP ticket

the current `ROLE` list doesn't include `Artist` and `Contributing_Artists` as listed on that metadata

src/pages/home/uploadSong/AddOwnerModal.tsx
dmkirshon: Agree with Trevor, and on board now with not disabling the button in this scenario and any form based settings. I believe the red warning borders and messages provide enough context for the user, especially since they won't be able to move forward without completing the indicated fields, I wouldn't say any of those errors could be missed in that scenario. If needed can bring up to design team if a polite toast message or informational bubble pops up saying please fill in the fields. As long as there's enough context for the user to understand what needs to be submitted than I think we are winning. Also agree on disabling in the "consent" scenario.

In this scenario I believe it's also important to properly notify the user of those two trigger option requirements, maybe a red box around both signifying a need for selection, I just don't think the words underneath the two boxes does it justice. Either way we should be consistent with types of user warnings we give for each input field type.

Bit of a rabbit hole, but my thought is that this modal or minting in general might require some level of first time tour or video informational clues because minting is a big deal and understanding all these pieces will be important for the user as well, to give a little more context of the page and not just hey! a field missing type of warning. Future Post MVP possibly.

escobarjonatan: From the convo above, i will be leaving it as is, in this case it will be enabled by default and then display the errors.

src/pages/home/library/Mint.tsx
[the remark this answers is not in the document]
escobarjonatan: given that we aren't sure what the end result would look like, we can add that later

src/pages/home/library/Mint.tsx
escobarjonatan: @dmkirshon Confirmed that a song can be minted without credits. will keep as is 🤷🏽‍♂️

dmkirshon: Sounds good, just seems a little strange to leave out for metadata unless the artist wants complete anonymity I guess.

PR #183, opened 2023-02-04
Add GetSongGenres API call used in the upload songs page.

cullenpu:
Demo:

https://user-images.githubusercontent.com/51892908/216791686-7b29986b-4f43-4b3d-a62a-31dd6e8ac5d7.mov

### DIFF
--- a/src/modules/content/slice.ts
+++ b/src/modules/content/slice.ts
@@ -3,38 +3,6 @@ import { ContentState } from "./types";
 
 const initialState: ContentState = {
   // TEMP: Data is mocked until API data is available
-  genres: [
-    "Alternative",
-    "Anime",
-    "Blues",
-    "Children's",
-    "Classical",
-    "Clasicalelle",
-    "Comedy",
-    "Commercial",
-    "Country",
-    "Dance",
-    "Electronic",
-    "Enka",
-    "Enko",
-    "Hip Hop",
-    "Indie Pop",
-    "Industrial",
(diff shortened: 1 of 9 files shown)


src/pages/home/uploadSong/UploadSong.tsx
```diff
@@ -24,8 +24,14 @@ import MintSong from "./MintSong";
...
 
-  const { genres } = useSelector(selectContent);
-  const isLoading = useSelector(selectSongIsLoading);
+  const { data = [] } = useGetSongGenresQuery();
+  const genres: string[] = data.filter((g) => {
+    // Filter out empty genre
+    if (g.length > 0) {
+      return g;
+    }
+  });
```
escobarjonatan, line 33: You can simplify this to:
`const genres = useGetSongGenresQuery().data.filter(genre => genre.length > 0);`
if you are not a fan of the one liners you can do this:

```
const { data = [] } = useGetSongGenresQuery();
const genres = data.filter(genre => genre.length > 0);

```
  
 A side note to the `g` param, as you iterate over things where you may not care too much about the single item, it makes it easier to read the code if you name it what it is that way you don't need to add the comment and it is self documented.

cullenpu: I switched to your two line suggestion, thanks!

escobarjonatan:
Closing as changes are no longer valid.

PR #184, opened 2023-02-09
Implementing password reset functionality

escobarjonatan:
[Figma mocks](https://www.figma.com/file/zrn01ZR2zFeLiKBebEYI1v/Artist-Portal---Final-Designs?node-id=2386%3A27569&t=HWkOlqNRuLYAI91o-0)

https://user-images.githubusercontent.com/11530120/217723307-2004771c-bfb5-48f5-91d9-54d99fbc8c0d.mov


src/modules/session/thunks.ts
```diff
@@ -72,6 +76,42 @@ export const createAccount = createAsyncThunk(
...
+        heading: "Password changed!",
+        message: "Login with the newly defined password.",
+        severity: "success",
+      })
+    );
+
+    setTimeout(() => {
+      window.location.pathname = "login";
+    }, 2000);
+  }
```
escobarjonatan, line 112: @scandycuz I wanted to get your thoughts on this. In the mocks it shows the redirect to login and then after, you show the toast. I tried different ways to do this but was unsuccessful, the only way that i think we can accomplish that is through sessionStorage and then dispatching the toast if content is present in sessionStorage in the login page. Do you have any other things we can try to accomplish what they want? should i go with the sessionStorage solution? or should i ask design team if they are okay with this?

dmkirshon: I'm also curious to see if there is another way around this without changing too much of the current structure. Would it be possible to add the `useNavigate` hook to the `Toast` component and then use dispatch to provide an optional navigation route to then display the message?

escobarjonatan: I can leave as is for now and when we merge the other PR we can make the change here, i dont want to copy/paste code around

src/pages/forgotPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const theme = useTheme();
+  const [maskPassword, setMaskPassword] = useState(true);
+  const {
+    isValid,
+    values: { confirmPassword, newPassword },
+  } = useFormikContext<FormikValues>();
+
+  const togglePasswordMask = () => {
+    setMaskPassword(!maskPassword);
+  };
```
dmkirshon, line 17: Is this considered a duplicate or just added as safety?` PasswordInputField.tsx` has the following as default:
```
  const togglePasswordMask = () => {
    setInternalMaskPassword(!internalMaskPassword);
  };
  ...
  
  onClick={ handlePressEndAdornment || togglePasswordMask }
```

src/pages/forgotPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const showEndAdornment = !!(newPassword || confirmPassword);
```
dmkirshon, line 19: This is just understanding for myself, I see that everything works correctly and the mask icon appears once a change occurs in one of the input text fields, but this is a `const` and I don't see anything in `PasswordInputField.tsx` that may change this value to trigger the `endAdornment`.

src/pages/forgotPassword/VerifyEmail.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+        <Button sx={ { mt: 2 } } type="submit">
+          Reset password
+        </Button>
+      </Box>
+
+      <Box my={ 4 } textAlign="center">
+        { showResendLink ? (
+          <Typography color="grey100" fontWeight={ 500 }>
+            Didn&apos;t received the email?
```
dmkirshon, line 71: Just small typo from copy 
`Didn't receive an email? Click here to resend`

src/pages/forgotPassword/VerifyEmail.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const handleEmailResend = () => {
+    setShowResendLink(false);
+
+    dispatch(sendVerificationEmail(email));
+
+    setTimeout(() => setShowResendLink(true), 10000);
+  };
+
+  return (
+    <Box
```
dmkirshon, line 26: I've seen other places with `Container` and others with `Box`, is there any subtle difference between the two in the top wrapper layer?

src/pages/forgotPassword/VerifyEmail.tsx
dmkirshon: Also the same on the Verify page for new users on `pages/signUp/Verification.tsx` if you make the change now instead of adding it to the ticket of copy changes before finishing up MVP.

src/pages/forgotPassword/ResetPassword.tsx
dmkirshon: Yea I do understand that portion I guess its the connection to this in `PassworInputField.tsx`:

```
endAdornment={
        showEndAdornment ? (
          <IconButton
            aria-label="Change password field visible status"
            onClick={ handlePressEndAdornment || togglePasswordMask }
          >
            <PasswordIcon sx={ { color: theme.colors.white } } />
          </IconButton>
        ) : undefined
      }
```
Where if showEndAdornment is set to false on empty strings and passed in, how does it update to become true, is it just wiffing my head as a basic react Component process, or is it a Formik trick in updating input fields and components

src/modules/session/thunks.ts
escobarjonatan: Thanks, implemented this in the latest push.

src/pages/forgotPassword/VerifyEmail.tsx
escobarjonatan: Good call, i fixed this in the latest push.

PR #190, opened 2023-02-27
Setting up initial iDenfy KYC verification

escobarjonatan:
Add the ability for a user to be able to verify themselves through iDenfy KYC services.

https://user-images.githubusercontent.com/11530120/221598470-d305fc0e-7e26-48a1-9e46-ea1b025a4f05.mov


src/api/idenfy/types.ts
```diff
@@ -0,0 +1,65 @@
...
+  readonly dateOfExpiry?: string | null;
+  readonly dateOfIssue?: string | null;
+  readonly documentNumber?: string | null;
+  readonly documents?: string[];
+  readonly errorUrl?: string;
+  readonly expiryTime?: number;
+  readonly externalRef?: string | null;
+  readonly firstName?: string;
+  readonly generateDigitString?: boolean;
+  readonly lastName?: string;
```
escobarjonatan, line 18: along with the required `clientId` that we send to them, we can also send the optional `firstName` `lastName` fields. This would ensure the person doing the verification session has the same name as the profile. Maybe we can do that post MVP?

src/common/mockConstants/mockSession.ts
```diff
@@ -4,4 +4,7 @@ import { mockProfile } from "./mockProfile";
 export const mockSession: SessionState = {
   isLoggedIn: true,
   profile: mockProfile,
+  idenfy: {
+    authToken: "",
+  },
```
escobarjonatan, line 9: I thought we should put the authToken in state so we dont trigger an api call to fetch a token every time they open/close the modal? We can set the expiry time of the session/token in the admin settings as well

src/elements/styled/Tooltip.tsx
```diff
@@ -0,0 +1,28 @@
+import { styled } from "@mui/material/styles";
```
escobarjonatan, line 1: Copied from the token sale

src/modules/session/thunks.ts
```diff
@@ -95,6 +97,27 @@ export const createAccount = createAsyncThunk(
...
+      idenfyApi.endpoints.requestVerificationToken.initiate(body)
+    );
+
+    if ("error" in idenfyTokenResponse) {
+      return;
+    }
+
+    dispatch(receiveIdenfyToken(idenfyTokenResponse.data.authToken));
+
+    dispatch(updateProfile({ verifiedStatus: "pending" }));
```
escobarjonatan, line 117: This is something that has been on my mind. We update status to "pending" after we initiate the verification session. if the user decides not to finish the process, it will remain in pending. They will still be able to create a new session if they login out/in or refresh. So technically this will still work but it would not be an accurate status. @scandycuz  thoughts?

src/modules/session/types.ts
```diff
@@ -14,6 +17,7 @@ export interface Profile {
   readonly pictureUrl: string;
   readonly role: string;
   readonly genre: string;
+  readonly verifiedStatus: "unverified" | "pending" | "verified";
```
escobarjonatan, line 20: I've thought about this a little more, FE should never update the verifiedStatus to "verified". Should we create 2 fields instead? @scandycuz thoughts?
`readonly isVerified: boolean;`
`verifiedStatus: "unverified" | "pending";`

src/pages/home/profile/IdenfyModal.tsx
```diff
@@ -0,0 +1,36 @@
+import { FunctionComponent } from "react";
+import { useSelector } from "react-redux";
+import { selectSession } from "modules/session";
+import { Modal } from "components";
```
escobarjonatan, line 4: I noticed that this modal that was brought in from the token sale is full screen. Are we going to keep our modals full screen or did we import the wrong one? @scandycuz

src/pages/home/profile/Profile.tsx
```diff
@@ -24,12 +31,30 @@ const Profile: FunctionComponent = () => {
...
+  const isPendingVerification = verifiedStatus === "pending";
+  const isVerified = verifiedStatus === "verified";
+
+  const handleCloseModal = () => {
+    setIsModalOpen(false);
+  };
+
+  const handleVerificationSession = () => {
+    if (!authToken) {
+      dispatch(requestVerificationToken({ clientId: id }));
```
escobarjonatan, line 54: If you are testing this out in your local, I would recommend commenting this line out after you get the authToken and hardcode that in. We have a limited amount of testing tokens and would like to use them once the BE work is done so that is more E2E.

src/modules/session/thunks.ts
escobarjonatan: @scandycuz After learning more about iDenfy and the process. The piece that i couldn't get around was what you last stated, currently there's no support for adding a callback on token request. 

However, it looks like we will be getting an automatic callback as soon as the user finishes their session. At that point BE will update the status to pending and then approved as soon as they get the second callback. 

We can configure the success/fail url and add a param like you suggested and then send a request to BE for the status, but at that point, why not go with the other solution to send BE a status request every 1 minute or so after triggering a session verification. I can see us having to send interval requests even with the query solution.

I updated the code to remove changing status for now though

src/modules/session/types.ts
escobarjonatan: You are correct, that change will be reflected in the next push

src/pages/home/profile/IdenfyModal.tsx
escobarjonatan: I will have a conversation with product bc i feel like the full screen is a better option for iframes as well to allow flexibility vs trying to fit a full page in 600px lol

escobarjonatan:
Reference discussion: https://github.com/projectNEWM/newm-artist-portal/pull/190#discussion_r1118901320

It's not letting me comment on the actual thread but Trevor and i had a slack discussion where we discussed option #1 more in depth and this is the path forward:

After the user finishes the process through iFrame or other device:
On success:
1. They should be redirected to success url w/ params`?idenfyVerification="pass"`.(configured through iDenfy console)
3. FE starts pinging BE services for 20 min(1 min interval) or until user status is verified, whichever comes first.

On fail:
1. They should be redirected to fail url w/ params`?idenfyVerification="fail"`.
2. We dispatch a toast message letting them know.

Before we do this, we need to implement redux-persist.

PR #193, opened 2023-03-02
Updating login/forgot password page to match the mocks

escobarjonatan:
https://user-images.githubusercontent.com/11530120/222338079-a0bca6f7-b5ab-4ec4-9d01-f563d6879b90.mov


src/App.css
```diff
@@ -3,15 +3,11 @@
...
 @property --angle {
-  syntax: "<angle>";
+  syntax: '<angle>';
   initial-value: 0deg;
   inherits: false;
 }
 
-.MuiSvgIcon-root {
-  color: #DC3CAA;
-}
```
escobarjonatan, line 13: I figured as we are making new changes and bringing in mui icons, we would rip the bandaid off and if you see an icon that is suppose to be this color, we can change it on the icon itself and not globally like this was doing. 
CC: @cullenpu @dmkirshon @scandycuz

dmkirshon (changes requested):
All looks good, just noticed there was no validation on the `VerifyEmail` page. Missing the `Invalid entry. Please check the code and try again.` 
Noticed the issue, where if you run through the reset password process, `VerifyEmail` page lets you enter any code and move forward, no code triggers a validation error. The next `ResetPassword` page provides a toast error when trying to change the password though if an invalid code is entered, but for a user if they don't copy paste might not run into issue if they fat finger it. Validation on the verify email code should stop all of that.

escobarjonatan:
@dmkirshon Good question, the reason why i didnt add that validation was because we dont have a way to validate the verification code itself, besides checking that it is a 6 digit number. The VC is only good for 1 time use and expires in 5 minutes for security reasons.
Atm, we are unable to accomplish what the mock shows. When working on the forgot password flow i brought up the idea of us potentially having a single page to submit the new password and verification code so that we would be able to validate, but since the signup flow has them as separate pages, they went for consistency, which is understandable. Maybe this is something we deal with on the QA phase to improve UX flow?

dmkirshon (approved):
LGTM, agree QA phase, maybe a better error message to inform incorrect/not valid verification code entered and a navigation to restart process.

PR #196, opened 2023-03-07
Adding redux-persist to keep track of verification start date

escobarjonatan:


src/modules/session/slice.ts
```diff
@@ -19,6 +19,7 @@ const initialState: SessionState = {
     genre: "",
     verificationStatus: "Unverified",
   },
+  verificationPingStartedAt: new Date().getTime(),
```
escobarjonatan, line 22: This will be undefined, i just added time right now for testing.

PR #198, opened 2023-03-09
Updating logic to request profile on success page, fixing postMessage

escobarjonatan:


src/components/idenfy/IdenfyModal.tsx
```diff
@@ -23,24 +23,21 @@ const IdenfyModal: FunctionComponent<IdenfyModalProps> = ({
...
 
-  /** Triggers onClose when user is verified and removes verification timer. */
-  if (verificationStatus === "Verified") {
-    const event = new Event("close");
-
+  /** Removes verification timer when verified. */
+  if (isVerified && verificationPingStartedAt) {
     dispatch(removeVerificationTimer());
-
-    onClose(event);
```
escobarjonatan, line 43: Removed the modal close event here when the user is verified bc it seems a bit aggressive and we can just let them close the modal, and when they do, they wont be able to open it again bc ✨magic ✨

PR #203, opened 2023-03-20
CU-8669nr1ah: select multiple moods for song

scandycuz:
Implements selecting multiple moods for a song.

### Also

- Switches to using only RTK Query for content API calls instead of also using Redux. Currently, the hook will always fetch new data when it's used in a component, however, I made a ticket to re-enable RTK Query caching so that data will only be re-fetched when necessary: https://app.clickup.com/24371401/v/b/q7r69-4208.

Makes various updates to upload/update song flow:

- Disables button when it is loading
- Updates `getAudioUploadUrl` request to use updated endpoint url
- Navigates to library page when uploading or updating a song completes

### DIFF
--- a/src/api/newm/api.ts
+++ b/src/api/newm/api.ts
@@ -13,6 +13,7 @@ const api = createApi({
   baseQuery: fetchBaseQueryWithReauth(baseQuery),
   // TODO: Update caching behavior for endpoints to avoid re-fetching data on
   // each mount (docs: https://redux-toolkit.js.org/rtk-query/usage/automated-refetching).
+  // ClickUp ticket: https://app.clickup.com/t/8669z53e4
   refetchOnMountOrArgChange: true,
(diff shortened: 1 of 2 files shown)


src/modules/song/thunks.ts
```diff
@@ -83,8 +81,8 @@ export const uploadSong = createAsyncThunk(
         body: JSON.stringify({ file: audioBinaryStr }),
       });
 
-      // fetch user's songs
-      dispatch(songApi.endpoints.getSongs.initiate());
+      // navigate to library page to view new song
+      history.push("/home/library");
```
escobarjonatan, line 85: lol i actually did this on my local branch as well, minus the comment. 
We are doing a getSongs fetch somewhere else too and i took that off, i'll explain in my PR but this change should be good

escobarjonatan (approved):

PR #204, opened 2023-03-20
Implements verifying profile from upload/update song flows

scandycuz:
- Link to iDenfy flow from upload/update song flows 
- Hide the "Verify profile" button from the upload/update song flows if the user is already verified

### Also

- Uses TypeScript enum for verification status, just to help ensure values are always valid

### DIFF
--- a/src/components/idenfy/IdenfyModal.tsx
+++ b/src/components/idenfy/IdenfyModal.tsx
@@ -16,16 +16,14 @@ const IdenfyModal: FunctionComponent = () => {
     Cookies.get("idenfyAuthToken")
   );
 
-  if (!idenfyAuthToken) {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -37,6 +38,7 @@ const App = () => {
           <Toast />
           <CssBaseline />
           <IdenfyPingUserStatus />
+          <IdenfyModal />
```
escobarjonatan, line 41: I get why you added this to the global app, but i think it might be an overkill. With this, it would run some logic on every single page load, even after the user is verified. I think we should only add the component where needed at the moment it is only in 2 places, even if it was in 5 i dont think the performance hit is worth putting it at the app level. Thoughts?

escobarjonatan (approved):

PR #208, opened 2023-03-25
Remove encoding audio files and adding temporary support for mp3 files

escobarjonatan:


src/components/UploadSong.tsx
```diff
@@ -59,11 +59,12 @@ const UploadSong: FunctionComponent<UploadSongProps> = ({
     [onChange, onBlur, onError]
   );
 
+  // TODO: Remove ".mp3" support after transcoded is fixed for supported files
   const { getRootProps, getInputProps, isDragActive } = useDropzone({
     onDrop: handleDrop,
     multiple: false,
     accept: {
-      audio: [".flac", ".fla", ".wav"],
+      audio: [".flac", ".fla", ".wav", ".mp3"],
```
escobarjonatan, line 67: @scandycuz It seems like currently there's no support for transcoding the rest of the file types. So that we can test the upload flow i've added .mp3 files, we can delete after they add the support for it. What do you think?

src/modules/song/thunks.ts
```diff
@@ -74,13 +74,12 @@ export const uploadSong = createAsyncThunk(
       if ("error" in audioUploadUrlResp) return;
 
       const { uploadUrl } = audioUploadUrlResp.data;
-      const audioBinaryStr = await getFileBinary(body.audio);
 
       // upload audio to AWS, song audioUrl will be updated after it's transcoded
       await fetch(uploadUrl, {
         method: "PUT",
         headers: { "Content-Type": "application/json" },
-        body: JSON.stringify({ file: audioBinaryStr }),
```
escobarjonatan, line 83: Encoding the audio file seems to increase the file size and it is not a requirement for the upload to process successfully, we can send it as is.

src/components/UploadSong.tsx
dmkirshon: I'm not on review or called out here, but I just tested with .wav on local and the upload and generation of stream url worked. Not sure if this part is needed anymore.

The only issue I had was upload processing on the [PR preview link](https://newm-artist-portal--pr208-hotfix-upload-song-e-k7w1g421.web.app/), both mp3 and wav files didn't produce a stream url.

escobarjonatan: >I'm not on review or called out here 
lol feedback is always welcome, i thought this was going to be merged before devops fixed the issue but they're freaky fast.

I was able to upload them fine locally... I will remove the mp3 as it is not longer needed

src/modules/song/thunks.ts
```diff
@@ -74,13 +74,12 @@ export const uploadSong = createAsyncThunk(
...
 
       const { uploadUrl } = audioUploadUrlResp.data;
-      const audioBinaryStr = await getFileBinary(body.audio);
 
       // upload audio to AWS, song audioUrl will be updated after it's transcoded
       await fetch(uploadUrl, {
         method: "PUT",
-        headers: { "Content-Type": "application/json" },
-        body: JSON.stringify({ file: audioBinaryStr }),
+        headers: { "Content-Type": "application/octet-stream" },
```
escobarjonatan, line 81: devops suggested to use this so that our flac and wav files get uploaded and encoded successfully.

src/components/UploadSong.tsx
```diff
@@ -63,7 +63,7 @@ const UploadSong: FunctionComponent<UploadSongProps> = ({
     onDrop: handleDrop,
     multiple: false,
     accept: {
-      audio: [".flac", ".fla", ".wav"],
+      audio: [".flac", ".wav"],
```
escobarjonatan, line 66: Removing fla to avoid confusion as suggested here: https://projectnewmgroup.slack.com/archives/C04P2C60JS1/p1680618743534499

PR #210, opened 2023-03-31
Adding more fields to the user profile and matching figma mocks

escobarjonatan:
- Adding new profile fields
- Match recent mocks
- Added initial logic for optional field text to show on text input, dropdown and text area

Note: This does not take care of the user image and banner, it is a separate ticket

image


src/pages/home/profile/Profile.tsx
```diff
@@ -161,127 +185,204 @@ const Profile: FunctionComponent = () => {
...
-                    "repeat(1, 1fr)",
-                    null,
-                    "repeat(2, 1fr)",
-                  ],
-                  rowGap: ["16px", null, "12px"],
-                  columnGap: [undefined, undefined, "20px"],
-                  maxWidth: [undefined, undefined, "700px"],
-                } }
+              <Box
+                sx={ { maxWidth: ["340px", "340px", "700px"], margin: "0 auto" } }
```
dmkirshon, line 192: The two text input rows start bunching up at the third breakpoint so I'm moving the breakpoint on the Settings page to make it look fluid.

```suggestion
                sx={ { maxWidth: { xs: "340px", lg: "700px" }, margin: "0 auto" } }
```

___
And same change needed for the flex directions below to keep it lined up.

`flexDirection: { xs: "column", lg: "row" },`

src/pages/home/profile/Profile.tsx
```diff
@@ -161,127 +185,204 @@ const Profile: FunctionComponent = () => {
...
+                  sx={ {
+                    columnGap: 2,
+                    flexDirection: [null, null, "row"],
+                    mt: 5,
+                    rowGap: 2,
+                  } }
+                >
+                  <Button
+                    disabled={ !dirty }
+                    onClick={ handleReset }
```
dmkirshon, line 361: I made this Button `type="reset"` on the Settings page, does that get handled differently than a  `handleReset` passed through Formik?

I also have this button `color="music"` per the Figma design.

src/pages/home/profile/Profile.tsx
escobarjonatan: I think this needs to be its own ticket where we do a find and replace all an test it everywhere. Otherwise, this PR will be bigger than it needs to. What do you think?

src/pages/home/profile/Profile.tsx
```diff
@@ -1,19 +1,26 @@
 import { FunctionComponent } from "react";
 import { useDispatch, useSelector } from "react-redux";
-import { Container, IconButton, Stack } from "@mui/material";
+import { Box, Container, IconButton, Stack } from "@mui/material";
 import HelpIcon from "@mui/icons-material/Help";
 import CheckCircleIcon from "@mui/icons-material/CheckCircle";
 import { Form, Formik, FormikValues } from "formik";
+import * as Yup from "yup";
+import { useGetSongsQuery } from "modules/song";
 import { Button, HorizontalLine, Tooltip, Typography } from "elements";
```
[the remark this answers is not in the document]
escobarjonatan: I think we should try to eliminate the use cases of our elements typography component bc i feel it adds confusion and if the only benefit is for us to be able to change colors easier we are talking about this:

`<Typography color="primary">text</Typography>`
vs 
`<Typography sx={{color: theme.palette.primary}}>text</Typography>`

The second one gives more flexibility and would remove the confusion of having 2 separate typography components. If you agree @scandycuz i can replace the element typography use here and bring in the mui element.

src/pages/home/profile/Profile.tsx
dmkirshon: Agree, as discussed before the artist-portal needs adjustments for mobile responsiveness, probably with some added figma designs, this could be a part of that epic or just a general UI clean up task. All for the separate ticket.

PR #211, opened 2023-04-01
Adding optional text to optional fields

escobarjonatan:
Started making this change as part of the profile update but figured i should do it as is own PR to keep it simple. I am sure in the future we will be making changes to what the switch, datepicker, image and other non common optional fields. I figured this was a start for that

### DIFF
--- a/src/elements/DropdownSelect.tsx
+++ b/src/elements/DropdownSelect.tsx
@@ -19,6 +19,7 @@ export interface DropdownSelectProps
   readonly errorMessage?: string;
   readonly handleChange?: (newValue: string) => void;
   readonly label: string;
+  readonly isOptional?: boolean;
   readonly name: string;
   readonly noResultsText?: string;
   readonly options: string[];
--- a/src/elements/TextInput.tsx
+++ b/src/elements/TextInput.tsx
@@ -75,7 +75,7 @@ export const TextInput: ForwardRefRenderFunction<
     startAdornment,
     endAdornment,
     disabled = false,
-    isOptional = false,
+    isOptional = true,
     tooltipText = "",
     widthType = "default",
     ...rest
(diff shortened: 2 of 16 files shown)


src/elements/DropdownSelect.tsx
```diff
@@ -19,6 +19,7 @@ export interface DropdownSelectProps
   readonly errorMessage?: string;
   readonly handleChange?: (newValue: string) => void;
   readonly label: string;
+  readonly isOptional?: boolean;
```
dmkirshon, line 22: Is this missing the add to the actual component or will that occur after the merge?

src/elements/TextInput.tsx
```diff
@@ -75,7 +75,7 @@ export const TextInput: ForwardRefRenderFunction<
     startAdornment,
     endAdornment,
     disabled = false,
-    isOptional = false,
+    isOptional = true,
```
dmkirshon, line 78: Just to make sure this change gets incorporated fully, I noticed the use with setting `TextInputField` and `PasswordInputField` with `isOptional` since it extends `TextInputProps`, will it add value as part of this change if you add a `false` default within the `PasswordInputField.tsx`?
Or do you see it being set within each individual page. My thought is that if there is a PasswordInputField on the page it will always be required, and by default `false` for `isOptional`. I understand that it may get convoluted per set defaults in your current structure.

With all that, if you follow current pattern, when this is merged in we will need to incorporate it as part of the Settings.tsx page as well.

dmkirshon (approved):
LGTM, just two comments about general pattern

src/elements/DropdownSelect.tsx
escobarjonatan: This will occur after the merge. I realized that the actual code that does the actual optional functionality is part of the other PR. 🥲

src/elements/TextInput.tsx
escobarjonatan: Very good point, it doesnt hurt to set the default to false in the password field, if we need to override it anywhere to optional we can pass is the true value

escobarjonatan:
@scandycuz this is what was decided by design, https://www.figma.com/file/zrn01ZR2zFeLiKBebEYI1v?node-id=1922:25058#402974192

image

PR #213, opened 2023-04-06
CU-8669z53e4: implement RTK Query caching functionality

scandycuz:
- Removes temporary work-around to always fetch new data and implements RTK Query caching tags so that data is only refetched when necessary.

### Also

- Replaces Redux functionality with RTK Query hooks where applicable.
- Adds an `asThunkHook` util function that allows using thunks as hooks so that it isn't necessary to store the loading state in Redux.

### DIFF
--- a/src/pages/createProfile/CreateProfile.tsx
+++ b/src/pages/createProfile/CreateProfile.tsx
@@ -1,6 +1,9 @@
 import { FunctionComponent } from "react";
 import { Box, Container, useTheme } from "@mui/material";
-import { ProfileFormValues, useUpdateInitialProfile } from "modules/session";
+import {
+  ProfileFormValues,
+  useUpdateInitialProfileThunk,
+} from "modules/session";
 import { WizardForm } from "components";
 import { commonYupValidation } from "common";
 import * as Yup from "yup";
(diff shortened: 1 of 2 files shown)


src/api/cloudinary/types.ts
```diff
@@ -6,7 +6,7 @@ export interface CloudinaryUploadParams extends CloudinaryUploadOptions {
   readonly api_key: string;
   readonly file: string | ArrayBuffer | null;
   readonly signature: string;
-  readonly timestamp: string;
+  readonly timestamp: number;
```
escobarjonatan, line 9: should this be `string | number` ?

src/pages/createProfile/CreateProfile.tsx
```diff
@@ -46,8 +39,8 @@ const CreateProfile: FunctionComponent = () => {
   /**
    * Submits the form when on the last route of the form.
    */
-  const handleSubmit = (values: FormikValues) => {
-    dispatch(updateInitialProfile({ ...values }));
+  const handleSubmit = (values: ProfileFormValues) => {
+    updateInitialProfile({ ...values });
```
escobarjonatan, line 43: This feels different, not sure if good or bad lol but i like the pattern

src/pages/home/library/SongInfo.tsx
```diff
@@ -47,32 +55,16 @@ const SongInfo = () => {
...
-    if (JSON.stringify(genres) !== JSON.stringify(values.genres)) {
-      updatedValues.genres = values.genres;
-    }
-    if (JSON.stringify(moods) !== JSON.stringify(values.moods)) {
-      updatedValues.moods = values.moods;
-    }
-    if (title !== values.title) {
-      updatedValues.title = values.title;
-    }
+    const updatedValues = getUpdatedValues(initialValues, values);
```
escobarjonatan, line 65: yea!!!

escobarjonatan (approved):

PR #219, opened 2023-04-16
CU-85zrrnnru | Changes required to support new fields from the upload song url endpoint

jayvaznewm:
Changes required to read the pre-signed post data and pass the fields back when uploading the song.

### DIFF
--- a/src/modules/song/thunks.ts
+++ b/src/modules/song/thunks.ts
@@ -53,13 +53,22 @@ export const uploadSong = createAsyncThunk(
 
       if ("error" in audioUploadUrlResp) return;
 
-      const { uploadUrl } = audioUploadUrlResp.data;
+      const { url: uploadUrl, fields } = audioUploadUrlResp.data;
 
-      // upload audio to AWS, song audioUrl will be updated after it's transcoded
+      // build a form with AWS presigned fields and upload audio to AWS
+      //  song audioUrl will be updated after it's transcoded
+      const formData = new FormData();
+      for(const key in fields) {
+        formData.append(key, fields[key]);
+      }
+      formData.append("file", body.audio);
       await fetch(uploadUrl, {
-        method: "PUT",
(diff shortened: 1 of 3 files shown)


src/modules/song/thunks.ts
```diff
@@ -53,13 +53,22 @@ export const uploadSong = createAsyncThunk(
...
+      formData.append("file", body.audio);
       await fetch(uploadUrl, {
-        method: "PUT",
-        headers: { "Content-Type": "application/octet-stream" },
-        body: body.audio,
+        method: "POST",
+        headers: {
+          "Content-Type": "application/octet-stream",
+          "ContentDisposition": `filename=${body.audio.name}`
+        },
```
escobarjonatan, line 70: Since we will be using `FormData`, i wanted to test out the `formData.getHeaders()` method so we can potentially get rid of the explicit octet content type and let the form decide. So updated code would look like:

```
      const { url: uploadUrl, fields } = audioUploadUrlResp.data;

      // build a form with AWS presigned fields and upload audio to AWS
      //  song audioUrl will be updated after it's transcoded
      const formData = new FormData();
      for(const key in fields) {
        formData.append(key, fields[key]);
      }
      
      formData.append("file", body.audio);
      
      const headers = new Headers({
        "Content-Disposition": `filename=${body.audio.name}`
      });
      
      await fetch(uploadUrl, {
        method: "POST",
        headers: Object.assign({}, headers, formData.getHeaders()),
        body: formData,
      });
```

Not sure if this would work but wanted to test it out. Let me know when you plan to merge the other changes so that i can test these and possibly merge around the same time

jayvaznewm: @escobarjonatan  - I incorporated portions of this, but `formData.getHeaders()` doesn't exist, however I was able to confirm that the correct Content type (`/audio`) is automatically applied, which I assume is being handled by the form.

escobarjonatan (approved):

PR #221, opened 2023-04-18
Hotfix: Upload song bugs

scandycuz:
Fixes the following bugs:

- Regression where artist agreement wasn't stored in Redux
- Bug where wizard form loading status wouldn't reset after `onSubmitStep` was called.

### DIFF
--- a/src/modules/song/thunks.ts
+++ b/src/modules/song/thunks.ts
@@ -9,6 +9,7 @@ import {
   UploadSongRequest,
 } from "./types";
 import { extendedApi as songApi } from "./api";
+import { receiveArtistAgreement } from "./slice";
 
 /**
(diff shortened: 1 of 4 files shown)


src/modules/song/thunks.ts
```diff
@@ -84,7 +85,11 @@ export const generateArtistAgreement = createAsyncThunk(
 
       if ("error" in artistAgreementResp) return;
 
-      callback();
+      dispatch(receiveArtistAgreement(artistAgreementResp.data.message));
+
+      if (typeof callback === "function") {
```
escobarjonatan, line 90: Good check here 👏🏽

escobarjonatan (approved):

PR #222, opened 2023-04-20
CU-866a4j7am: switch from video.js libary to hls.js library

scandycuz:
Switches to the hls.js library instead of the video.js library.

The hls.js library seems to have more descriptive error messages than the video.js library, which were helpful to debug the CORS issue that was preventing songs from being played. It is also a bit smaller.

### Notes

- When testing, songs that wouldn't play before due to the CORS issue may still not play, which looks to be due to CloudFront caching behavior. If newly uploaded songs continue to have CORS issues, then it will need to be looked into further, however, it seems to be resolved.


escobarjonatan (approved):
🔥🔥🔥

PR #223, opened 2023-04-21
Create Settings Page

dmkirshon:
## Acceptance criteria
 - The user is able to update their password and delete their account from the settings page.
 - The save button is disabled if the user hasn't made any changes the the form (`Formik dirty  value is false), enabled if any changes have been made.
 - Terms of Service is linked for user navigation
 - Logout button is available for users to logout of Artist-Portal

## Additional features
- If user is logged in with Social OAuth (Google, Facebook), the change password identifies that for the user and the cancel/save buttons are hidden
- If current password is same as new password a validation statement appears to make user aware

[Settings_Page_Demo.webm](https://user-images.githubusercontent.com/86050631/233672286-c6bc86f5-dcf7-4a41-a9d2-42000838b674.webm)


src/modules/session/thunks.ts
```diff
@@ -164,6 +167,32 @@ export const createAccount = createAsyncThunk(
...
+  async (body: DeleteAccountRequest, { dispatch }) => {
+    const deleteAccountResponse = await dispatch(
+      sessionApi.endpoints.deleteAccount.initiate(body)
+    );
+
+    if ("error" in deleteAccountResponse) {
+      return;
+    }
+
+    dispatch(logOut());
```
dmkirshon, line 184: Doesn't completely remove state when using OAuth social accounts, after account deletion (no refresh or closing browser) the user is able to select their social and login again, old profile details persist

This is a strange edge case with cache/state where refresh or closing the browser fixes the issue

Note: When using the same OAuth login after delete account and refresh or close browser, user would get stuck in a loop of the create-profile steps, but again this is an edge case if the user decides to create another account with same socials right after deleting their first one. (username and password works normally in this case with no looping of create-profile)

src/pages/home/settings/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,126 @@
...
+                    by typing &ldquo;YES&rdquo; below.
+                  </Typography>
+                </Stack>
+                <Stack
+                  sx={ {
+                    display: "flex",
+                    justifyContent: "start",
+                  } }
+                >
+                  <TextInputField name="confirmationStatement"></TextInputField>
```
escobarjonatan, line 87: Not a required change but this can be self enclosing: 
`<TextInputField name="confirmationStatement" />`

src/pages/home/settings/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,126 @@
...
+                    alignItems: "end",
+                    columnGap: 1.5,
+                    pt: 1.5,
+                  } }
+                >
+                  <Button
+                    variant="secondary"
+                    color="music"
+                    width="compact"
+                    onClick={ handleCloseDialog }
```
escobarjonatan, line 104: In a state where `isLoading` is true, would you want to disable the cancel button so they dont exit early, making them think they canceled the process? just some thoughts 🤔

escobarjonatan (approved):

src/pages/home/settings/DeleteAccountDialog.tsx
dmkirshon: I think the thought flew through my head, but didn't add it on first pass. Will add it before I merge.

src/pages/home/settings/DeleteAccountDialog.tsx
dmkirshon: Will fix this in addition to the comment below before merge.

dmkirshon: I remember why I went with the extra wrapper Stack, the TextInputField uses a base HTML element and not a MUI variant. Was on a flow of keeping it within the `sx` pattern within the component. 

Will follow suit though and remove the extra Stack and use the style prop to handle this one.

escobarjonatan: Oh I wasn't talking about the stack wrap. I was referring to this:
<TextInputField name="confirmationStatement" />

Vs

<TextInputField name="confirmationStatement"></TextInputField>

Sorry for the confusion

escobarjonatan (approved):

PR #236, opened 2023-05-04
CU-866a6t2u1: upload and mint song with collaborators

scandycuz:
Saves song collaborators when uploading a song.

### DIFF
--- a/src/api/newm/api.ts
+++ b/src/api/newm/api.ts
@@ -11,7 +11,7 @@ const baseQuery = fetchBaseQuery({
 
 const api = createApi({
   reducerPath: "newmApi",
-  tagTypes: [Tags.Profile, Tags.Song, Tags.Collaborator],
+  tagTypes: [Tags.Profile, Tags.Song, Tags.Collaborator, Tags.Collaboration],
   baseQuery: fetchBaseQueryWithReauth(baseQuery),
   endpoints: () => ({}),
 });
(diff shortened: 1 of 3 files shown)


src/api/newm/api.ts
```diff
@@ -11,7 +11,7 @@ const baseQuery = fetchBaseQuery({
 
 const api = createApi({
   reducerPath: "newmApi",
-  tagTypes: [Tags.Profile, Tags.Song],
+  tagTypes: [Tags.Profile, Tags.Song, Tags.Collaborator, Tags.Collaboration],
```
escobarjonatan, line 14: I added `Collaborator` on my PR so we might have a small merge conflict

src/modules/song/api.ts
```diff
@@ -45,7 +49,7 @@ export const extendedApi = api.injectEndpoints({
       async onQueryStarted(body, { dispatch, queryFulfilled }) {
         try {
           await queryFulfilled;
-        } catch ({ error }) {
+        } catch (err) {
```
escobarjonatan, line 52: can we do `error` vs `err`. I try to be explicit when possible. i've seen places where e is used for event and error and it causes confusion as to what it is. In this case err cannot be confused with event lol but still 🤷🏽‍♂️

escobarjonatan (approved):

PR #238, opened 2023-05-07
Updating logic to process accepted status of stream token agreement

escobarjonatan:


src/modules/song/thunks.ts
```diff
@@ -112,6 +112,21 @@ export const uploadSong = createAsyncThunk(
             },
           })
         );
+
+        if ("error" in generateArtistAgreementResponse) return;
+
+        if (
+          generateArtistAgreementResponse.meta.requestStatus === "fulfilled"
+        ) {
```
escobarjonatan, line 120: I dont think this check is needed if `"error" in generateArtistAgreementResponse` is true, it will return and not run this code. Just adding it for now to make sure it only runs "fulfilled"

src/pages/home/owners/OwnersTable.tsx
```diff
@@ -125,7 +123,7 @@ export default function OwnersTable({
               index
             ) => (
               <TableRow key={ id || index }>
-                <TableCell>
+                <TableCell sx={ { display: { xs: "none", sm: "table-cell" } } }>
```
escobarjonatan, line 126: Realized that the name is not always going to be available. The song count and email will be though so I changed the table to display those 2 on smaller screens

dmkirshon:
So to confirm/reiterate what was discussed previously. This change will update the backend when that the user agrees to the stream token agreement and submits for minting. correct?

I've run it locally and following through the path under Upload A Song it does update auto-magically to Pending, but if I try and do the minting process after upload, on the Library page through the songs Edit Song page, it does not update the backend, no redux action is sent as viewed by redux-logger. Is Edit Song page part of this task?

I believe it's because the Edit Song page doesn't have the same flow as Upload A Song.

On another side note I'm seeing another redux action call for `generateArtistAgreement` from both Edit Song page and Upload A Song after the Request to Mint submit button is clicked, not sure why that is happening.

dmkirshon (changes requested):
So to confirm/reiterate what was discussed previously. This change will update the backend when that the user agrees to the stream token agreement and submits for minting. correct?

I've run it locally and following through the path under Upload A Song it does update auto-magically to Pending, but if I try and do the minting process after upload, on the Library page through the songs Edit Song page, it does not update the backend, no redux action is sent as viewed by redux-logger. 

Is Edit Song page part of this task? I believe it's because the Edit Song page doesn't have the same flow as Upload A Song.

On another side note I'm seeing another redux action call for `generateArtistAgreement` from both Edit Song page and Upload A Song after the Request to Mint submit button is clicked, not sure why that is happening.

escobarjonatan:
> So to confirm/reiterate what was discussed previously. This change will update the backend when that the user agrees to the stream token agreement and submits for minting. correct?
> 
> I've run it locally and following through the path under Upload A Song it does update auto-magically to Pending, but if I try and do the minting process after upload, on the Library page through the songs Edit Song page, it does not update the backend, no redux action is sent as viewed by redux-logger.
> 
> Is Edit Song page part of this task? I believe it's because the Edit Song page doesn't have the same flow as Upload A Song.
> 
> On another side note I'm seeing another redux action call for `generateArtistAgreement` from both Edit Song page and Upload A Song after the Request to Mint submit button is clicked, not sure why that is happening.

This change will not technically update the BE, it will user agreement accept/decline status of token agreement before minting.

Minting process after upload is currently being worked on by @scandycuz

Edit song is not part of this task

`generateArtistAgreement` being called twice is expected, it should be called once after user clicks on next when uploading song, then it would be called again on upload and start minting with `saved` and `songId` in the payload to save the agreement in the bucket.

dmkirshon (approved):

PR #240, opened 2023-05-16
Edit and view collaborators

scandycuz:
Allows for editing and viewing collaborators for uploaded songs.

### DIFF
--- a/src/components/idenfy/IdenfyModal.tsx
+++ b/src/components/idenfy/IdenfyModal.tsx
@@ -17,8 +17,8 @@ const IdenfyModal: FunctionComponent = () => {
   );
 
   const handleClose = useCallback(() => {
-    dispatch(setIsIdenfyModalOpen(false));
-  }, [dispatch]);
+    if (isOpen) dispatch(setIsIdenfyModalOpen(false));
+  }, [dispatch, isOpen]);
 
   /** Listens for modal close message. */
   useEffect(() => {
--- a/src/components/idenfy/IdenfyPingUserStatus.tsx
+++ b/src/components/idenfy/IdenfyPingUserStatus.tsx
@@ -14,8 +14,12 @@ import {
 
 const IdenfyModal: FunctionComponent = () => {
   const dispatch = useDispatch();
+  const { isLoggedIn } = useSelector(selectSession);
 
-  const { data: { verificationStatus } = emptyProfile } = useGetProfileQuery();
+  const { data: { verificationStatus } = emptyProfile } = useGetProfileQuery(
+    undefined,
+    { skip: !isLoggedIn }
+  );
 
   const { verificationPingStartedAt } = useSelector(selectSession);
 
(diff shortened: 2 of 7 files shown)


src/modules/song/thunks.ts
```diff
@@ -179,7 +185,101 @@ export const patchSong = createAsyncThunk(
...
+        }
+      }
+
+      if (body.title && body.artistName) {
+        await dispatch(
+          generateArtistAgreement({
+            body: {
+              artistName: body.artistName,
+              companyName: body.companyName,
+              save: true,
```
escobarjonatan, line 275: This should be `saved: true,`

src/api/lambda/types.ts
```diff
@@ -1,8 +1,8 @@
 export interface GenerateArtistAgreementBody {
   readonly songName: string;
-  readonly companyName: string;
+  readonly companyName?: string;
   readonly artistName: string;
-  readonly stageName: string;
+  readonly stageName?: string;
   readonly songId?: string;
   readonly save?: boolean;
```
escobarjonatan, line 7: I think after you merge master, you should see this as `saved` and not `save`

escobarjonatan (changes requested):
Minor detail on the GenerateArtistAgreementBody implementation, everything else looks good.

dmkirshon (changes requested):
Had some general questions, some of this might just be other tasks within the collaborator feature set:

1. When I had a song already submitted for minting, I was able to add/change owners/credits to the song. However strange things start happening with ownership when submitting Request Minting second time. I've seen percentages are off (over 100%) for a case of removing owners. Might be a different task as this is ownership and not for collaborators specifically.

Steps to recreate over 100% ownership:
a. Add song and request mint with two additional owners, both credited
b. Go into the Song Edit page and removed both owners but keep them on credits.
c. Resubmit Request Minting
d. See on Song Edit page under minting tab that the owners did not get removed and over 100% listed.

2. Also I had a song in `Pending` mint status, with only my ownership and credit (awaiting my own acceptance as collaborator on a song, which seems strange), which had a static display of 100% ownership shares. I was able to add owners and credits though but the % was static for my ownership so couldn't change their numbers to move on. Attempting to resubmit Request Mint with new creditors did lead to a toast message `An error occured while adding a collaborator`.
This might be an old data issue causing strange states, where the Minting status not being `Undistributed` because of logic path of PR#238. 
Will probably be a few kinks while these Collaborator features get merged, but I just wanted to provide the comment that I would have expected `Add New Owner` button to be greyed out or not visible if the status isn't Undistributed. 

3. Should I be able to click Next on the Minting tab if none of the owners or credits have changed?

4. Might be a side discussion with design or a code change, but it's strange to see myself under Collaborators, maybe should be denoted differently somehow?

escobarjonatan (approved):

dmkirshon (approved):
LGTM

PR #243, opened 2023-05-22
Skip data fetch when user is not logged in and updating iDenfy modal …

escobarjonatan:
…to trigger a close if it's opened

### DIFF
--- a/src/components/idenfy/IdenfyModal.tsx
+++ b/src/components/idenfy/IdenfyModal.tsx
@@ -17,8 +17,8 @@ const IdenfyModal: FunctionComponent = () => {
   );
 
   const handleClose = useCallback(() => {
-    dispatch(setIsIdenfyModalOpen(false));
-  }, [dispatch]);
+    if (isOpen) dispatch(setIsIdenfyModalOpen(false));
+  }, [dispatch, isOpen]);
 
   /** Listens for modal close message. */
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


src/components/idenfy/IdenfyModal.tsx
```diff
@@ -17,8 +17,8 @@ const IdenfyModal: FunctionComponent = () => {
   );
 
   const handleClose = useCallback(() => {
-    dispatch(setIsIdenfyModalOpen(false));
-  }, [dispatch]);
+    if (isOpen) dispatch(setIsIdenfyModalOpen(false));
```
escobarjonatan, line 20: Previously, if you hit the `escape` it would trigger a close of modal bc it is an app level component. I've added this check so that it doesn't trigger it if it's closed.

src/components/idenfy/IdenfyPingUserStatus.tsx
```diff
@@ -14,8 +14,12 @@ import {
 
 const IdenfyModal: FunctionComponent = () => {
   const dispatch = useDispatch();
+  const { isLoggedIn } = useSelector(selectSession);
 
-  const { data: { verificationStatus } = emptyProfile } = useGetProfileQuery();
+  const { data: { verificationStatus } = emptyProfile } = useGetProfileQuery(
+    undefined,
+    { skip: !isLoggedIn }
```
escobarjonatan, line 21: This was the culprit of us getting that error toast on login.
Reason, since this is an app level component it triggers a fetch profile data even if user is not logged in, we need to skip it if that is the case.

PR #244, opened 2023-05-22
Updating createAccount to updated POST endpoint and response

escobarjonatan:


src/pages/home/library/MintSong.tsx
```diff
@@ -1,4 +1,4 @@
-import { Box, Button as MUIButton, Stack } from "@mui/material";
+import { AlertTitle, Box, Button as MUIButton, Stack } from "@mui/material";
```
escobarjonatan, line 1: Inserting minor change as this is a small enough PR to update copy/colors to match mocks

PR #246, opened 2023-05-30
CU-866ab6dz7: Add featured artist collaborator

scandycuz:
Allows for adding featured artist collaborators.

### Note

- This work doesn't include displaying the profile image and name for an approved collaborator.

### DIFF
--- a/src/modules/song/types.ts
+++ b/src/modules/song/types.ts
@@ -231,6 +231,7 @@ export interface CreateCollaborationRequest {
   readonly role: string;
   readonly royaltyRate?: number;
   readonly credited: boolean;
+  readonly featured: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


src/modules/song/types.ts
```diff
@@ -185,6 +195,9 @@ export interface GetCollaborationsRequest {
   readonly statuses?: CollaborationStatus[];
 }
 
+/**
+ * Collaboration as stored in the NEWM API.
```
escobarjonatan, line 199: I like this idea, to add comments to the interfaces that belong to an endpoint, it would make it easier to update in the future 🤷🏽‍♂️

escobarjonatan (approved):

PR #249, opened 2023-06-01
Updating copy for adding owners and tooltip

escobarjonatan:
image


src/components/minting/SelectCoCreators.tsx
```diff
@@ -137,6 +139,30 @@ const FormContent: FunctionComponent<FormContentProps> = ({
         </Button>
       </Stack>
 
+      <Stack mt={ 1.5 } flexDirection="row">
+        <Typography variant="subtitle2">
+          For every additional collaborator who will receive royalties, the fee
+          to complete the minting process will increase by 1.5 ADA.
```
dmkirshon, line 145: Saw a little discussion back and forth on the CU task and Slack but wanted to get clarity on this. This 1.5 ADA amount isn't set right? Do we have a scheme for these values to be updated? Or is that only happening on a task for integrating the Minting backend process?

src/components/minting/SelectCoCreators.tsx
```diff
@@ -137,6 +139,30 @@ const FormContent: FunctionComponent<FormContentProps> = ({
...
+        </Typography>
+        <Tooltip
+          title={
+            "This cost is increased with each additional collaborator because " +
+            "this ADA needs to travel with each portion of stream tokens in order " +
+            "to complete the split, which is an added extra cost for each transaction."
+          }
+        >
+          <IconButton sx={ { padding: 0 } }>
+            <HelpIcon
```
dmkirshon, line 155: Just a note, but can be addressed during on the mobile responsive effort, the position of the help icon looks out of place when below the md (900px) sizing.

dmkirshon (approved):
LGTM, just two comments for clarity

src/components/minting/SelectCoCreators.tsx
escobarjonatan: uhm, i didnt think it looked that bad, but added 8 px spacing 🫡

src/components/minting/SelectCoCreators.tsx
escobarjonatan: Correct, it's not set. They are still trying to figure it out and how the collaborators affect the pricing from what i recall. Either way this should be small enough to implement at a later stage once we know the math part

src/components/minting/SelectCoCreators.tsx
dmkirshon: Just looked cramped on my screen 😅 looks good!

PR #255, opened 2023-06-10
Applying custom hook to scroll to the first error in a form

escobarjonatan:
https://github.com/projectNEWM/newm-artist-portal/assets/11530120/aae44c29-96b8-4944-8318-283f2981da2e


dmkirshon (approved):
LGTM, only other one I can see having this is change password on Settings.tsx page but that is a single section vs the many fields on these other pages.

PR #258, opened 2023-06-12
Updating table footer to match mocks

escobarjonatan:
Update table footer to match mocks across library/owners/wallet
Update password reset requirements with social auth
Users who signed up with socials should now be able to set a password and login through same social or password 
Update song validation requirements(file size, sample rate, duration) to meet EVEARA criteria.


src/pages/home/settings/Settings.tsx
```diff
@@ -32,22 +34,47 @@ const Settings: FunctionComponent = () => {
...
       then: Yup.string().required("Must match new password"),
-    }),
-  });
+    }
+  );
+
+  /**
+   * Defines the validation schema based on the user's login method.
+   * If the user is logged in using social login, only the new password and the confirmation password are required.
+   * If the user logged in using email and password, the current password is also required.
```
dmkirshon, line 64: Should we provide some type of copy for the social login situation? Seems strange to have the password fields without any other indication that it will set up their email login... This doesn't need email verification as well?

dmkirshon (approved):
LGTM, just one comment to ponder on Social Logins

src/pages/home/settings/Settings.tsx
escobarjonatan: I think this is one of those things where we can get feedback on that and add if needed. You are correct, this doesn't need email verification. It was discussed that for MVP we allow them to do this and provide a different approach post-MVP.

PR #293, opened 2023-06-27
Error handling for advanced details isrc field

escobarjonatan:
https://github.com/projectNEWM/newm-artist-portal/assets/11530120/7b96dca5-9986-4353-a30b-eb2d7981eec3


dmkirshon (approved):
LGTM

This might be a future UX fix. When trying to modify one of the characters in the `ISRC` input field it causes the entire input to shift and sends the cursor to the end of the input 🤔

PR #295, opened 2023-06-27
Connect Advanced Details field values to Upload A Song form flow

dmkirshon:
This PR incorporates the Advanced Details field values into the new song upload for the Upload A Song flow.

Notes:
 - Language value is also added to initial values
 - The Date field only includes `releaseDate` (Scheduled Release Date), the Original Release Date field may get removed and doesn't currently have a backend endpoint parameter
 - User IPI isn't integrated yet so this only sends an IPI value if the user adds one in the advanced details field, this also means collaborator/creditor IPIs will need to be integrated separately 

CU#862k0bf3n


src/pages/home/uploadSong/UploadSong.tsx
```diff
@@ -44,6 +44,15 @@ const UploadSong: FunctionComponent = () => {
...
     description: "",
+    language: undefined,
+    copyrights: undefined,
+    barcodeType: undefined,
+    barcodeNumber: undefined,
+    isrc: undefined,
+    iswc: undefined,
+    ipis: undefined,
+    userIpi: "",
+    releaseDate: undefined,
```
dmkirshon, line 55: Added these as placeholders to note all the initialValues, they're listed as undefined so they aren't sent through to the endpoint if left empty. Is it useful to include them to note which values might be touched or should they be left off? Should we modify the other optional fields to undefined? The only one that will become prefilled is the userIpi.

dmkirshon: That was my thinking as well but leaning on the other end. Wanted the second opinion and it is redundant to have those values since when excluded they do still present as `undefined`. There are also other indicators to what fields/values are present. I will clean it up to removed the optional parameters.

dmkirshon: Just double checked a different scenario of just adding and then removing data from the optional Advanced Details page fields. When the fields are touched they end up as string values which doesn't play nice with the endpoint. We will need to do checks on values that are empty strings and replace them with undefined.

PR #297, opened 2023-06-30
CU-866agx28q: updates wallet portfolio to use song data

scandycuz:
- Sets portfolio transaction data to users songs (will be updated with actual royalty data post-MVP).

### Also

- Stops playing uploaded song when upload song component unmounts


dmkirshon (approved):
LGTM, a separate issue but still got the scroll bars... the `overflow: scroll` in `SongRoyaltiesList.tsx` is causing it but I think we might be able to overrides these across the Artist Portal with a CSS scroll style change to make it a little more consistent across browsers.

Screenshot from 2023-06-30 16-44-49

PR #299, opened 2023-07-07
Navigate to song after updating

scandycuz:
Navigates to a song after it is updated

### Also

- Uses song ID in url to fetch song data
- Clears RTK query cache when logging out


dmkirshon (changes requested):
Two comments for review:
1. Navigation issue with edit-song page after save
2. Security of edit-song page with songId url

PR #309, opened 2023-07-13
Miscellaneous fixes from team session meeting part 3

escobarjonatan:
- Add validation for song form genre field to limit to 5, not as many as they want
- Investigate frontend rounding error possible for song length conversion
- Default sort song library by date descending
- Front-end ping for song minting status every 60 seconds or so (in addition to pinging for encoded song url)
- Style scrollbars to match app theme

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -1,26 +1,23 @@
 /* Firefox support */
-:root {
+.modify-scrollbar :root {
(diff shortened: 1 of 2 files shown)


dmkirshon (approved):
LGTM, two non-blocking comments included

PR #311, opened 2023-07-16
Misc updates

scandycuz:
- Ensure upload form is always on initial step on refresh (this can be removed when form data is persisted)
- Formats checkboxes so text is aligned with checkbox
- Updates audio upload component to prevent selecting non-audio files

<table>
<tr>
<td>
Screen Shot 2023-07-15 at 8 35 52 PM
</td>
<td>
Screen Shot 2023-07-15 at 8 31 33 PM
</td>
</tr>
</table>

### DIFF
--- a/src/components/UploadSong.tsx
+++ b/src/components/UploadSong.tsx
@@ -111,7 +111,7 @@ const UploadSong: FunctionComponent<UploadSongProps> = ({
     onDrop: handleDrop,
     multiple: false,
     accept: {
-      audio: [".flac", ".wav"],
+      "audio/*": [".flac", ".wav"],
     },
   });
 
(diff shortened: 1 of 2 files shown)


src/elements/Checkbox.tsx
```diff
@@ -54,12 +61,21 @@ const Checkbox: ForwardRefRenderFunction<HTMLInputElement, CheckboxProps> = (
...
+        { typeof label === "string" ? (
+          <Typography
+            variant="subtitle1"
+            sx={ { color: theme.colors.white, fontSize: 12 } }
+          >
+            { label }
+          </Typography>
+        ) : (
+          label
+        ) }
```
dmkirshon, line 74: Is this just a general pattern to provide a default instead of typing the parameter in one direction? Is there reason to leave the `ConfirmContract.tsx` in a Typography wrapper?

dmkirshon (approved):
LGTM, just one comment on `Typography` wrappers for `Checkbox.tsx`

PR #328, opened 2023-08-01
CU-866aq6305: prompt user around wallet address updates trevor scandalios

scandycuz:
- Prompts user to change wallet env and refresh page if wallet env doesn't match app env
- Prompts user whether they would like to update their currently saved address after connecting a wallet


dmkirshon (approved):
Overall LGTM, I have a few general comments, questions and edge cases:
- If I already have a wallet address connected and disconnect and reconnect that same wallet, should I still get a modal asking if I want to change my wallet address?
- Is the ability to pick a different wallet address from the same wallet extension a post-mvp feature?
- After switching between wallets (connect & disconnect between two wallets), none of the dialogs/modals pop up when connecting again to the existing wallets again, including the `do you want to change wallet address?` modal. (Gets resolved on refresh)
- I'm sure it's just in these specific scenarios but there were a lot of modals popping up (the invite modal and the wallet modals) that might overwhelm the UX. Need to review the layering/display before release.

PR #331, opened 2023-08-02
CU-866ak27n1: displays wallet balance and address

scandycuz:
Screen Shot 2023-08-02 at 12 23 09 AM

### DIFF
--- a/src/components/DisconnectWalletButton.tsx
+++ b/src/components/DisconnectWalletButton.tsx
@@ -8,25 +8,19 @@ import ArrowDropDownIcon from "@mui/icons-material/ArrowDropDown";
 import ContentCopyIcon from "@mui/icons-material/ContentCopy";
 import DoneIcon from "@mui/icons-material/Done";
 import LogoutIcon from "@mui/icons-material/Logout";
-import {
-  emptyProfile,
-  selectSession,
-  useGetProfileQuery,
-} from "modules/session";
 import { FunctionComponent, useEffect, useRef, useState } from "react";
 import { setIsConnectWalletModalOpen } from "modules/ui";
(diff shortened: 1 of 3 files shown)


src/components/DisconnectWalletButton.tsx
dmkirshon, line 1: Is this ready for review? Having a couple issues using the new button:
1. I'm having issues disconnecting from a wallet after immediatly connecting (requires a refresh), works fine if refreshed after connecting and then disconnecting
2. the env modal keeps popping up every 3 seconds (when it is supposed to, just very aggressive, on top of having to refresh to disconnect)
3. In certain instances the modal for updating the recently connected wallet runs into a loop when you click No and keeps popping back up
4. The successfully connnected toast message keeps popping up even when closing it out

dmkirshon: Deleting my node modules and lock files as @escobarjonatan advised seemed to do the trick for my 1-3 issues and the 4th item seems to be gone as well.

I think I'm running into a few other instances of the modals not popping up at the right time and other finnicky things, but I'm not sure if it's because I'm intentionally clicking things to the point where data is stale or some issues with my environment.

The other item I currently see is when I click the `Connect Wallet` button and try to add my wallet, my wallet details load into the component when the dialog to change addresses pops up and even if I hit `No` the wallet still shows connected. 
Would the expectations be to not load the wallet details until the user clicks `Yes`, or in a general case where a purge of all address info is done on clicking disconnect so the change in address modal doesn't display because the address was manually removed? In that case and just for future It will be important to display their wallet address to their song minting details as well.

Screenshot from 2023-08-03 11-07-33

PR #333, opened 2023-08-08
Create initial edit song commit to combine with upload song flow

escobarjonatan:
I still have some work left to do but wanted to get initial PR to get some eyes on it.

TODO:
- Determine what to do with song/stream url if should edit (probably a separate ticket as it needs feedback from product/BE, potentially post MVP and for now we would just hide that field)

https://github.com/projectNEWM/newm-artist-portal/assets/11530120/31cdb807-6d1e-4f75-83d9-b880f8760b92

### DIFF
--- a/src/components/PlaySong.tsx
+++ b/src/components/PlaySong.tsx
@@ -2,7 +2,7 @@ import PlayArrowIcon from "@mui/icons-material/PlayArrow";
 import StopIcon from "@mui/icons-material/Stop";
 import { FunctionComponent, useMemo, useState } from "react";
 import { useGetSongQuery, useHlsJs } from "modules/song";
-import { Typography } from "@mui/material";
+import { Stack, Typography } from "@mui/material";
 import IconMessage from "./IconMessage";
 
 interface PlaySongProps {
@@ -28,7 +28,7 @@ const PlaySong: FunctionComponent<PlaySongProps> = ({ id }) => {
   if (isLoading) return null;
 
   return song ? (
-    <>
+    <Stack sx={ { cursor: "pointer", width: "100%", height: "100%" } }>
       { isSongPlaying ? (
         <IconMessage
           icon={ <StopIcon /> }
@@ -42,7 +42,7 @@ const PlaySong: FunctionComponent<PlaySongProps> = ({ id }) => {
           onClick={ () => playSong(song) }
         />
       ) }
-    </>
+    </Stack>
   ) : (
     <Typography>Unable to play song</Typography>
   );


src/pages/home/library/EditSong.tsx
```diff
@@ -127,46 +289,58 @@ const EditSong = () => {
...
-
-      <TabPanel value={ tab } index={ 0 }>
-        <SongInfo />
-      </TabPanel>
-      <TabPanel value={ tab } index={ 1 }>
-        <MintSong />
-      </TabPanel>
-      <TabPanel value={ tab } index={ 2 }>
-        Marketplace content goes here
-      </TabPanel>
```
dmkirshon, line 169: 1. Is this a temporary removal, to be discussed with design/product as well? The current layout does seem a little strange as a one to one copy from the Upload Song Flow. 
2. I think some copy changes can help direct the user better in this edit flow, mainly for the first page `Basic Details`. Adding additional words like _Update Song File_ might be the minimum needed for MVP.
3. I think the UPLOAD A SONG heading gave the Upload Song Flow more context in that area, same as the previous tab structure of INFO/MINTING, might want to find a way to include EDIT/MINTING in the top heading for general context?

src/components/PlaySong.tsx
dmkirshon, line 1: Just wanted to double check if you ran into this, might need to open a separate bug ticket for it, the audio playback seems to be broken on the library page and edit song page. Play and stop are functioning but no audio output. No errors showing in console. Looks like an issue with `useHlsJs` as the `UploadSong` component is still working.

src/pages/home/uploadSong/BasicSongDetails.tsx
```diff
@@ -293,7 +324,7 @@ const BasicSongDetails: FunctionComponent = () => {
                 : "default"
             }
           >
-            { values.isMinting ? "Next" : "Upload" }
+            { values.isMinting ? "Next" : isInEditMode ? "Save" : "Upload" }
```
dmkirshon, line 327: Is the intention that even if I don't modify any data on the `BasicSongDetails` page, when I click save, it will still "save", navigate me to the library page, and show a toast saying "Updated song information"?

src/pages/home/uploadSong/BasicSongDetails.tsx
```diff
@@ -112,11 +122,32 @@ const BasicSongDetails: FunctionComponent = () => {
...
-          <UploadSongField name="audio" />
+              <SolidOutline
+                sx={ {
+                  display: "flex",
+                  justifyContent: "center",
+                  alignItems: "center",
+                  flexGrow: 1,
+                } }
+              >
+                <PlaySong id={ songId || "" } />
```
dmkirshon, line 139: Not a major item right now, unless a quick fix, but the responsive view of this box spans the width of the page above ~430px

dmkirshon (approved):
LGTM, had three items for review but otherwise see nothing blocking this PR.

PR #342, opened 2023-08-23
Add bottom padding to pages as needed and fix a minor wallet table issue

escobarjonatan:
- Add padding to the bottom of pages where needed when accessing it through smaller screens
- Fix table issue with wallet, where if the height of the window was smaller than the available space to display at least one row, it would fetch for 0 or -1, made it so it renders at least 1 row


dmkirshon (approved):
LGTM, I believe a larger conversation should happen soon about general responsiveness across the portal

PR #345, opened 2023-08-25
Enable submit buttons by default

escobarjonatan:


src/pages/home/profile/Profile.tsx
```diff
@@ -494,7 +494,6 @@ const Profile: FunctionComponent = () => {
                     </Button>
 
                     <Button
-                      disabled={ !dirty }
```
dmkirshon, line 497: This one produces an action and actually sends a query, even if data wasn't updated. Wanted to note the difference between the others being blocked by validation. Don't think it needs to be updated at this point.

Might be reasonable to add some copy in the future in the context of "no changes were made"

PR #351, opened 2023-09-02
User was redirected as soon as minting status changed. Added a check to not redirect if it was loading

escobarjonatan:
https://github.com/projectNEWM/newm-artist-portal/assets/11530120/29bfcae3-6681-4513-b4fc-45d6900c4dd7

### DIFF
--- a/src/components/PlaySongAdvanced.tsx
+++ b/src/components/PlaySongAdvanced.tsx
@@ -17,7 +17,7 @@ const PlaySongAdvanced: FunctionComponent<PlaySongAdvancedProps> = ({
   onPlayPause,
   song,
 }) => {
-  return song ? (
+  return song && song.streamUrl ? (
     <Stack
       sx={ {
         alignItems: "center",


src/pages/home/library/EditSong.tsx
```diff
@@ -201,43 +203,33 @@ const EditSong: FunctionComponent = () => {
...
+      ...values,
+      isMinting: false,
+      shouldRedirect: false,
+    };
 
-      navigate("advanced-details");
-    } else {
-      await handleSubmit(values, helpers);
+    if (step === "basic-details") {
+      patchValues.isMinting = !values.isMinting;
```
dmkirshon, line 217: I believe this is causing the patchSong thunk to execute the body.isMinting conditional 
`      if (body.isMinting) {` within `src/modules/song/thunks.ts` on line 418, which dispatches unnecessarily for edit song "basic details" updates, e.g. `generateArtistAgreement` is triggered for a basic save.

I think it would be fixed with just removing the line since it is false by default and the flow will still work, but if you see that it is needed within the logic just adding an additional statement or add an `else` for `body.shouldRedirect` would work: 
```     if (body.isMinting && !body.shouldRedirect) {```

**Review by dmkirshon — CHANGES_REQUESTED** [review]
Looks good overall, one point to fix/clean up and another general question for BE maybe

File: src/pages/home/owners/Songs.tsx [hunk 2]
```diff
@@ -47,9 +55,27 @@ const Songs: FunctionComponent = () => {
...
+          currentPlayingSongId: undefined,
+        }));
+      },
+      onSongEnded: () => {
+        setPlayerState((prevState) => ({
+          ...prevState,
+          isReadyToPlay: false,
+          currentPlayingSongId: undefined,
+        }));
+      },
```
dmkirshon, line 78: Can all of `playerState`/`hlsJsParams` Memo/`useEffects` be refactored into a custom hook and then referenced in each of these components, `PlaySong.tsx`/`InvitesTable.tsx`/`Songs.tsx` or does each component require custom tweaks right now?

dmkirshon (commented):
Reviewed the new material incorporating the song updates, had a few questions to follow up:

1. Do you have the local server set up to properly play the songs on local? Wanted to make sure the songs were coming through properly before updating this to garage. I might ask for assistance to set that up as I'm not able to play the songs currently on local. I haven't attempted to run through the document on my machine yet, got a little overwhelmed from the instruction set.
2. My internet connection might be slow for 'attempting' (see comment 1) to fetch some of the data, so I noticed a notable lag from selecting the play button to seeing the stop button on some of the library songs. Not sure if it will do the same on garage or you see the delay as well.
3. See comment below, refactor possible to consolidate the playback functionality?

Between the cookie security updates and the backend url updates I just wanted to be on the same page for your code changes.

dmkirshon (approved):
LGTM, all comments addressed

PR #353, opened 2023-09-03
CU-86a0g8nwc: adds composition and phonographic copyright fields

scandycuz:
Adds multiple fields for copyright.

### Also

- Disables Redux logging by default, since it can be noisy with RTK Query
- Resolves bug where all form fields would be touched after first wizard step submission

### DIFF
--- a/src/pages/home/uploadSong/AdvancedSongDetails.tsx
+++ b/src/pages/home/uploadSong/AdvancedSongDetails.tsx
@@ -3,6 +3,7 @@ import { useFormikContext } from "formik";
 import { Box, Link, Stack } from "@mui/material";
 import {
   MIN_DISTRIBUTION_TIME,
+  NEWM_ARTIST_PORTAL_FAQ_URL,
   NONE_OPTION,
   scrollToError,
   useWindowDimensions,
@@ -157,7 +158,7 @@ const AdvancedSongDetails = () => {
               owned by the artist and/or record label. If you are not the
               copyright holder of the sound recording, please review{ " " }
               <Link
-                href="https://newm.io/artists-faq"
+                href={ NEWM_ARTIST_PORTAL_FAQ_URL }
                 rel="noopener noreferrer"
                 target="_blank"
               >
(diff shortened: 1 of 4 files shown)


src/pages/home/uploadSong/AdvancedSongDetails.tsx
```diff
@@ -129,12 +143,54 @@ const AdvancedSongDetails = () => {
...
+          copyrightType="phonographic"
+          isOptional={ false }
+          tooltipText={
+            <span>
+              The copyright in a sound recording covers the recording itself (it
+              does not cover the music or lyrics of the song). It is typically
+              owned by the artist and/or record label. If you are not the
+              copyright holder of the sound recording, please review{ " " }
+              <Link
+                href="https://newm.io/artists-faq"
```
dmkirshon, line 160: 1. Created a `constants.ts` file so we don't have to jump across multiple files for external link changes
2. This link needs to be updated per one of the go-live-testing bug proposals, will be a quick change as part of that bug once the url is identified.
3. Might need to update to `NEWM_STUDIO` or just remove the `ARTISIT_PORTAL` portion of the naming convention completely if not relevant

https://github.com/projectNEWM/newm-artist-portal/blob/c62f1a9929ba131c3f14a41e3d0822b346eda832/src/common/constants.ts#L11

src/pages/home/uploadSong/UploadSong.tsx
```diff
@@ -177,13 +180,11 @@ const UploadSong: FunctionComponent = () => {
           rootPath="home/upload-song"
           isProgressStepperVisible={ true }
           enableReinitialize={ true }
-          validateOnBlur={ false }
           routes={ [
             {
               element: <BasicSongDetails />,
               path: "",
               progressStepTitle: "Basic details",
-              navigateOnSubmitStep: false,
```
dmkirshon, line 186: Currently navigating on Upload Only (No Mint) after progress bar to `advanced-details` instead of 'Library' page. I believe that is why this was set as false and the `navigate("advanced-details");` was in the conditional above at line 96

dmkirshon (changes requested):
New format for copyright is looking good, few changes requested to fix issues encountered

dmkirshon (approved):
LGTM, all comments addressed :+1:

PR #358, opened 2023-09-12
CU-86a0r3vj4: Move audio tests outside of Yup validation

dmkirshon:
Added:
- Custom error messages to handle audio tests on drag/drop

Removed:
- Yup validation for Audio, except for required

Bugs Noted:
- Large files over 100MB+ tend to crash Chrome when added, need to look into chunking bufferArray (see example txt attached)
Load times also take time for large files, noticeable delay of loading not allowing interaction with site, might require a loading UX
- Drag and drops have issues
   - issues with drag drop .mp3, error doesn't appear
   - issues with drag drop once audio file already added
- Noticed issue with Image drop, accepted non 1:1 ratio image and displayed error
- Once fields are already touched, Touching Image drop does not trigger Audio Required test, visa versa does trigger

### DIFF
--- a/src/common/formUtils.ts
+++ b/src/common/formUtils.ts
@@ -47,37 +47,6 @@ const includesGenres = (
   return hasValidGenre;
 };
 
-const createAudioBuffer = async (value: File) => {
-  const audioContext = new (window.AudioContext || window.AudioContext)();
-  const arrayBuffer = await value.arrayBuffer();
-
-  return audioContext.decodeAudioData(arrayBuffer);
-};
-
-const createAsyncAudioTest = (
-  testFn: (audioBuffer: AudioBuffer) => boolean
-) => {
-  return async (value: File) => {
-    if (!value) return false;
-
-    const audioBuffer = await createAudioBuffer(value);
-
-    return testFn(audioBuffer);
-  };
-};
-
-const isFileSizeValid = (value: File) => {
-  if (!value) return false;
-
-  const fileSizeInMB = value.size / (1024 * 1024);
-  const fileSizeInGB = value.size / (1024 * 1024 * 1024);
-
-  return (
-    fileSizeInMB >= AUDIO_MIN_FILE_SIZE_MB &&
-    fileSizeInGB <= AUDIO_MAX_FILE_SIZE_GB
-  );
-};
-
 const BARCODE_CONFIG: Record<BarcodeType | "DEFAULT", BarcodeConfig> = {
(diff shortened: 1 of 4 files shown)


dmkirshon:
Task linked: [CU-86a0mr633 Error sometimes incorrectly displays briefly after selecting genre when uploading song](https://app.clickup.com/t/86a0mr633)

src/common/formUtils.ts
```diff
@@ -47,37 +47,6 @@ const includesGenres = (
...
-
-    return testFn(audioBuffer);
-  };
-};
-
-const isFileSizeValid = (value: File) => {
-  if (!value) return false;
-
-  const fileSizeInMB = value.size / (1024 * 1024);
-  const fileSizeInGB = value.size / (1024 * 1024 * 1024);
```
dmkirshon, line 73: Thinking about switching this and album max value to 1000 instead of 1024 factor, as files are typically in bytes and Megabytes/Gigabytes (MB/GB) not Memibytes/Gebibytes (MiB/GiB). Do others agree?

src/components/UploadSong.tsx
```diff
@@ -42,18 +43,42 @@ interface SongProgressOverlayProps {
...
+  const createAudioBuffer = async (value: File) => {
+    const audioContext = new (window.AudioContext || window.AudioContext)();
+    const arrayBuffer = await value.arrayBuffer();
+
+    return audioContext.decodeAudioData(arrayBuffer);
+  };
+
+  const isFileSizeValid = (value: File) => {
+    const fileSizeInMB = value.size / (1024 * 1024);
+    const fileSizeInGB = value.size / (1024 * 1024 * 1024);
```
dmkirshon, line 74: Thinking about switching this and album max value to 1000 instead of 1024 factor, as files are typically in bytes and Megabytes/Gigabytes (MB/GB) not Mebibytes/Gebibytes (MiB/GiB). Do others agree?

dmkirshon:
Addressed issues and refactored to use `onError`, in addition migrated coverArt validation away from Yup for consistency/efficiency of file validations

dmkirshon:
> Nice, LGTM. Just to confirm, have the existing bugs mentioned in the PR description (other than crashing for really large files) been resolved? They seem to not occur when I tested the branch locally.

Re-tested and looks like the errors on drag and drop are working now, the second drag and drop issue is minor but still persisting. I will do a full write up for all of them and add them to the backlog.

PR #369, opened 2023-09-21
feat: Add archived to delete song endpoint

escobarjonatan:


cristhianescobar (approved):

dmkirshon (changes requested):
LGTM, just need to remove the collab reset when the upload song errors out. That way BE can keep track of collab issues as well.

I tested the removed code on my end and the collab email didn't go through and collabs didn't show up in my invite modal after "delete"/Archive. :+1: 

Remove:
https://github.com/projectNEWM/newm-artist-portal/blob/6a3cd8d608a91231caa2d2fb24a439036e2dc061/src/modules/song/thunks.ts#L255-L268

dmkirshon (approved):

PR #380, opened 2023-09-25
feat: Add query to fetch songs from wallet

escobarjonatan:
image


dmkirshon (approved):
LGTM! Only update/improvement I might consider, and let Product/Design know if needed, is updating the list/page when a new wallet is connected. Otherwise currently need to refresh the page to update the list of songs to the newly added wallet.

PR #390, opened 2023-10-10
feat: Added Sentry for error logging purposes

escobarjonatan:
Along with this setup, i also added "Sentry" slack app addon, you will see all new issues and resolved issues go through there. Created `#sentry-issues-web` to track production environment issues to reduce the noise. Alerts and notifications can be modified as we see fit.


dmkirshon (commented):
I will need to run through a bit of a tutorial for Sentry since I've never used it. Otherwise, found two minor things:

 1. Extra space in production string?
 2. Not seen in current Master branch but when loading this branch the text area input boxes on `Upload A Song` and `Profile` aren't styled correctly... I'm not sure how the changes you made affect this, but wanted to double check if you are seeing it on your side as well.
 
image

dmkirshon (approved):
LGTM

PR #399, opened 2023-10-21
Lock down name after verified, send all outlets when related fie…

escobarjonatan:
…lds change, update library play action

### DIFF
--- a/src/pages/home/library/SongList.tsx
+++ b/src/pages/home/library/SongList.tsx
@@ -207,6 +207,19 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
     event.stopPropagation();
   };
 
+  const handleRowClick = (
+    event: MouseEvent<HTMLButtonElement | HTMLTableRowElement>,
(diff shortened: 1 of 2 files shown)


src/pages/home/library/SongList.tsx
```diff
@@ -288,7 +301,13 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
...
               <TableRow
                 onClick={
-                  isSongStale ? undefined : () => handleSongPlayPause(song)
+                  isSongStale
+                    ? undefined
+                    : (event) =>
+                        handleRowClick(event, {
+                          songId: song.id,
+                          hasStartedMintingProcess,
+                        })
```
dmkirshon, line 310: Not a required change but to improve the readability of this would it help to refactor the isSongStale into the handleRowClick function or have a helper function so event is implicitly passed?

dmkirshon (approved):
LGTM, two refactoring questions but nothing blocking.

PR #410, opened 2023-11-17
Add DSP Pricing Plan feature

dmkirshon:
Creates new Distribution Service Provider (DSP) pricing plan popup window (`Dialog`) from CU-86a1cktby. This also integrates part of CU-86a1ckn3y: Toggling Mint Song opens DSP pricing plan. 

- Missing the ability to save users DSP plan choice to the backend. Blocked by backend changes required(CU-86a1ckyxj). Will create a new branch to handle saving users choice once backend endpoint is available.

[demo_NEWM-Studio_DSP-Pricing-Plan-Dialog.webm](https://github.com/projectNEWM/newm-artist-portal/assets/86050631/0a153c26-04e8-4775-b0ac-4b3ae77f1851)


src/assets/images/LeafFillIcon.tsx
```diff
@@ -0,0 +1,25 @@
+import { createSvgIcon } from "@mui/material";
+
+const LeafFillIcon = createSvgIcon(
```
dmkirshon, line 3: Couldn't find MUI icons that would fit the Figma design but looked into MUI icons a little more. I found this nice conversion/wrapping option that we should probably follow through and refactor all our previous custom icons to maintain consistency. 

https://mui.com/material-ui/icons/#createsvgicon

For our Figma svgs it must be noted that we need to set the `width` and `height` to `24px` to allow for the sizing to match up with MUI icons but otherwise I don't see any other issues or further customization needed. They act exactly like MUI icons and `sx` is available to change style, font-size and color of icons as needed.

src/components/dspPricing/PricingPlanOption.tsx
```diff
@@ -0,0 +1,172 @@
...
+            </Stack>
+          )) }
+        </Stack>
+
+        <Divider sx={ { width: "70%" } } color={ theme.colors.grey400 } />
+
+        { buttonType === "primary" ? (
+          <Button
+            onClick={ () => {
+              dispatch(setIsArtistPricePlanSelected(true));
```
dmkirshon, line 151: I used redux to save the state for now, but this will need to be replaced by the backend endpoint once the server change has been merged.

src/components/dspPricing/PricingPlansDialog.tsx
```diff
@@ -0,0 +1,95 @@
...
+};
+
+interface PricingPlansDialogProps {
+  readonly handleClose: () => void;
+  readonly open: boolean;
+}
+
+const PricingPlansDialog = ({ handleClose, open }: PricingPlansDialogProps) => {
+  return (
+    <Dialog onClose={ handleClose } open={ open } fullScreen sx={ { m: 5 } }>
```
dmkirshon, line 27: I'm going to ask this to Design/PM as well to get consistent nomenclature across our frontend implementations, but what is our definition of `Dialog` vs `Modal` vs whatever `Toast` style pop ups are? The CU task and other references just note this as a "Popup" and I feel we should get it straight across all parties.

It seems like this should be a `Modal` since it takes up the entire screen and doesn't let the user interact with anything else on the page. However, to match the Figma design UI I needed to use Dialog with some customizations.

src/components/dspPricing/pricingPlanData.json
```diff
@@ -0,0 +1,214 @@
+{
```
dmkirshon, line 1: I used a `json` file to capture all the details listed within the Figma design. 

1. I didn't see us using json for any other occasion so please let me know if there is a better or more code base consistent way to implement this data set.
2. I will double check with the Backend team to see if any of these details might come to the Frontend via an endpoint. From my understanding it is only the option to save the users DSP choice, but pricing from an endpoint might make sense here as well.

src/theme/theme.ts
```diff
@@ -78,6 +79,7 @@ declare module "@mui/material/styles" {
       grey500?: string;
       grey600?: string;
       grey700?: string;
+      lightGreen?: string;
```
dmkirshon, line 82: I added this as it was a new color seen in the DSP Pricing Plan Figma design but I need to have the conversation with Design to double check consistency across the design and nomenclature.

src/theme/theme.ts
```diff
@@ -277,6 +280,13 @@ const theme = createTheme({
       fontSize: "80px",
       lineHeight: "96px",
     },
+    h2: {
```
dmkirshon, line 283: I added an `h2` but I'm not sure this is what Design intended, it matched up with the font choice being used in the DSP Pricing Plan. 

However, there is a large conversation with Design that I want to start as the font's probably should consider accessibility and currently in the design it jumps around between `h2` to `h4` without consideration of hierarchy, only because the `fontFamily` and `fontSize` matches what is shown in Figma. My question would also be: can we get labeled style names in Figma so there isn't so much variation between text styles and we can match our theme template in a more consistent manner.

escobarjonatan (approved):

src/common/hooks.ts
```diff
@@ -110,3 +117,33 @@ export const usePrevious = <T>(value: T) => {
...
+ * @example
+ * useEffectAfterMount(() => {
+ *   // do something on next update, not including initial render
+ * }, []);
+ *
+ * @param cb the callback to run during the next component update
+ * @param dependencies the dependencies to watch for changes
+ */
+export const useEffectAfterMount = (
+  cb: EffectCallback,
```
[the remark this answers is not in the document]
dmkirshon: I will refactor this to make it clearer :+1:

src/modules/ui/slice.ts
```diff
@@ -17,6 +17,7 @@ const initialState: UIState = {
     message: "",
     isConfirmationRequired: false,
   },
+  isArtistPricePlanSelected: false,
```
scandycuz: It would make more sense to keep this in the `session` module, as it's related more to the user than just the UI.

We could wait until the back-end is updated first though, since where the data is stored will likely be determined by where the data comes from.

src/components/dspPricing/pricingPlanData.json
scandycuz: It might be better to include this in the assets directory, so that if json data is used somewhere else, they will all be stored in the same place. Eventually, we'd want to use some sort of strings file to contain all of the copy for the app, in order to making translating everything to different languages easier, but this works for now.

src/theme/theme.ts
scandycuz: Yeah, it would be ideal if design had a standard set of font variations that we could replicate in our theme.

scandycuz (commented):
LGTM, but had some comments.

src/modules/ui/slice.ts
dmkirshon: Makes a lot more sense to have in `session`, it's a pretty minor lift so I will look into moving it over just to keep everything logical

PR #411, opened 2023-11-19
feat: Add coverRemixSample toggle to the upload flow

escobarjonatan:
https://github.com/projectNEWM/newm-artist-portal/assets/11530120/4653362e-683b-423b-b135-e68e4f8f6aa7

### DIFF
--- a/src/components/minting/CoverRemixSample.tsx
+++ b/src/components/minting/CoverRemixSample.tsx
@@ -18,13 +18,13 @@ export const CoverRemixSample: FunctionComponent<CoverRemixSampleProps> = ({
   return (
     <SwitchInputField
       disabled={ disabled }
-      name="coverRemixSample"
+      name="isCoverRemixSample"
       title={
         "Is this song a cover, remix, mixtape, mashup, " +
         "or contain samples and/or any part of the intellectual property of another work?"
       }
     >
-      { values.coverRemixSample && (
+      { values.isCoverRemixSample && (
         <>
           <HorizontalLine mt={ 2 } />
 
--- a/src/components/minting/CoverRemixSample.tsx
+++ b/src/components/minting/CoverRemixSample.tsx
@@ -1,0 +1,64 @@
+import { FunctionComponent } from "react";
+import { Link, Typography } from "@mui/material";
+import { useFormikContext } from "formik";
+import { NEWM_STUDIO_FAQ_URL, NEWM_SUPPORT_EMAIL } from "common";
+import { SwitchInputField } from "components";
+import { Alert, HorizontalLine } from "elements";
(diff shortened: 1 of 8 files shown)


src/components/minting/CoverRemixSample.tsx
```diff
@@ -0,0 +1,64 @@
...
+              <Link
+                href={ `mailto:${NEWM_SUPPORT_EMAIL}` }
+                sx={ {
+                  color: "yellow",
+                  textDecorationColor: "yellow",
+                } }
+              >
+                support@newm.io
+              </Link>{ " " }
+              to proceed. See{ " " }
```
dmkirshon: This is more on the copy but this sounds like they can't even continue in the upload process before sending their license/documentation. Is that the intention? If they can submit/upload the song but need to do it in addition to finishing the mint process it probably should say something like "to have the audio file distributed/minted" or "after completing the upload process". Just feel like it needs more context and is a little ambiguous in the current state.

Also the "selected yes" above is strange in this context, might even advice that to be updated to "Due to this audio file containing cove/samples/mixes, ..."

src/modules/song/types.ts
```diff
@@ -92,6 +92,7 @@ export interface PostSongRequest {
   readonly ipis?: ReadonlyArray<string>;
   readonly releaseDate?: string;
   readonly publicationDate?: string;
+  readonly coverRemixSample?: boolean;
```
dmkirshon, line 95: I think this would benefit to be changed to `isCoverRemixSample` just to be consistent for `boolean` values.

dmkirshon (approved):
LGTM, just a copy question and a possible refactor.

src/modules/song/types.ts
escobarjonatan: This is how it is defined in the BE for post and get, I wanted to keep it consistent throughout so will keep it as is to not get confused on one vs the other.

scandycuz: I agree with David here, we should use the `is` prefix for booleans, even if the back-end doesn't. This could be interpreted in the code as being some sort of file as apposed to a boolean.

scandycuz (commented):
LGTM, just had one comment.

src/modules/song/types.ts
escobarjonatan: i'll update it

src/components/minting/CoverRemixSample.tsx
escobarjonatan: followed up with product and updated the copy

escobarjonatan:
Closing this in favor of the new https://github.com/projectNEWM/newm-web/pull/430

PR #431, opened 2023-12-04
Feature/dsp-pricing-plan

dmkirshon:
Adds the DSP Pricing Plan feature to NEWM Studio. The following tasks/components will need to be completed before merging the feature into the master branch:
- [X] [CU-86a1cktby: Create the DSP Pricing popup window](https://app.clickup.com/t/86a1cktby)
- [X] [CU-86a1ckn3y: Mint Song toggle button to trigger DSP pricing popup window](https://app.clickup.com/t/86a1ckn3y)
- [x] [CU-86a1ckn3y: Integrate endpoint to store DSP pricing plan user selection](https://app.clickup.com/t/86a1ckn3y)
- [X] [CU-86a1cktby: Integrate minting pricing endpoint into DSP pricing plan dialog](https://app.clickup.com/t/86a1cktby)


newmskywalker:
Task linked: [CU-86a1ckn3y Turning the Mint Song toggle button on should trigger the DSP Pricing popup window.](https://app.clickup.com/t/86a1ckn3y)

newmskywalker:
Task linked: [CU-86a1cktby Create the DSP Pricing popup window (within the Upload a Song form).](https://app.clickup.com/t/86a1cktby)

apps/studio/src/components/dspPricing/PricingPlansDialog.tsx
```diff
@@ -0,0 +1,123 @@
...
+                }
+                handleOptionClick={handleClose}
+                key={pricingPlan.id}
+                planIcon={{
+                  iconPxSize: ICON_SIZE,
+                  iconElement: PRICING_PLAN_ICON[pricingPlan.id],
+                }}
+                pricing={
+                  dspPriceUsd && pricingPlan.id === "artist"
+                    ? `$${Number(dspPriceUsd).toFixed(2)}/RELEASE`
```
dmkirshon: Should I refactor to use the `formatPriceToDecimal` once merged?

apps/studio/src/components/dspPricing/PricingPlansDialog.tsx
```diff
@@ -0,0 +1,137 @@
...
+          </Typography>
+          <Typography variant="h3">WHAT SUITS YOU BEST?</Typography>
+        </Stack>
+        <Stack
+          sx={{
+            alignContent: "center",
+            display: "flex",
+            flexDirection: "row",
+            gap: 4,
+            [theme.breakpoints.down("xl")]: {
```
scandycuz: The DSP pricing plan modal was vertically stacked on my laptop, even though the screen isn't that narrow. I'd change the theme's breakpoints to:

```
breakpoints: {
  values: {
    xs: 0,
    sm: 600,
    md: 960,
    lg: 1280,
    xl: 1600,
  },
},
```

and then change the responsive styling to change to a column on a `"lg"` screen and below, instead of `"xl"`. 

The responsive screen sizes were changed awhile ago to work better for the sample sales, but the sizes above are the Material UI defaults and probably work better.

Screen Shot 2023-12-07 at 7 40 17 PM

scandycuz (changes requested):
LGTM, but had a comment about the responsive sizing that I think would be good to address.

apps/studio/src/components/dspPricing/PricingPlansDialog.tsx
```diff
@@ -0,0 +1,137 @@
...
+  artist: (
+    <SeedlingFillIcon sx={{ color: theme.colors.music, fontSize: ICON_SIZE }} />
+  ),
+  artistPlus: (
+    <StarFillIcon sx={{ color: theme.colors.music, fontSize: ICON_SIZE }} />
+  ),
+};
+
+interface PricingPlansDialogProps {
+  readonly handleClose: () => void;
```
scandycuz: Just a syntax issue, but it's common practice to prefix function props with `on` and then have handlers inside of a component prefixed with `handle`. E.g. `<Example onClick={handleClick} />`.

Here's an article I found about it: https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364. It's mainly just helpful to avoid naming conflicts. E.g.

```
const Example = ({ onClose }) => {
  const handleClose = () => {
    // custom logic
  
    onClose()
  }
 
  ...
}
```
vs something like
```
const Example = ({ handleClose }) => {
  const _handleClose = () => {
    // custom logic
  
    handleClose()
  }
 
  ...
}
```

Also makes it easy to quickly know what is a prop.

apps/studio/src/components/dspPricing/PricingPlanOption.tsx
```diff
@@ -0,0 +1,185 @@
+import { Check, Close } from "@mui/icons-material";
+import { Box, Divider, Stack } from "@mui/material";
+import { Button, Typography } from "@newm-web/elements";
```
escobarjonatan: Typography should be imported from mui 🙃

apps/studio/src/components/dspPricing/PricingPlansDialog.tsx
```diff
@@ -0,0 +1,137 @@
+import { Box, IconButton, Stack } from "@mui/material";
+import { Dialog, Typography } from "@newm-web/elements";
```
escobarjonatan: import `Typography` from mui if possible

apps/studio/src/components/dspPricing/PricingPlansDialog.tsx
escobarjonatan: i think so, doesn't hurt. also we have somewhat similar solutions, i chose `parseFloat` over `Number` like you did here because in case we get `"1.233abcs"`, mine should output `"1.23"` vs in your case it would output to `NaN`.... small diff but think parseFloat is favored here in an edge case like this 🤷🏽‍♂️

escobarjonatan (approved):

apps/studio/src/components/dspPricing/PricingPlansDialog.tsx
dmkirshon: I believe I set it so conservatively partly because of those breakpoints but also due to how the text was bunching up. I will make these breakpoint changes, make sure the panels remain horizontal per your recommendations, and tell product to watch out for discrepancies across the app since it will affect other responsive views.

apps/studio/src/components/dspPricing/PricingPlansDialog.tsx
dmkirshon: Will create a refactor PR and include this.

scandycuz (approved):

PR #442, opened 2023-12-07
Add price summary with song estimates before mint process begins

escobarjonatan:
https://github.com/projectNEWM/newm-web/assets/11530120/811fd81d-1b4e-4a41-92a3-7dc2dddb5cdc


newmskywalker:
Task linked: [CU-86a1fcjwd Implement Payment Summary dialog as part of the Distribute & Mint user flow.](https://app.clickup.com/t/86a1fcjwd)

apps/studio/src/common/constants.ts
```diff
@@ -1,9 +1,3 @@
-/**
- * NEWM Fees
- */
-export const COLLABORATOR_FEE_IN_ADA = 1.3;
-export const MINTING_FEE_IN_ADA = 6;
-
```
escobarjonatan, line 6: Removed in favor of the song estimate query

scandycuz (approved):

packages/utils/src/lib/string.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Formats a price string to a specified number of decimal places.
+ * If the input price is undefined or cannot be parsed as a number,
+ * it defaults to 0. The function returns the formatted price as a string.
+ *
+ * @param {string} [price] - The price to be formatted. Optional.
+ * @param {number} [decimals=2] - The number of decimal places to format the price to. Defaults to 2.
+ * @returns {string} - The formatted price.
+ */
+export const formatPriceToDecimal = (price?: string, decimals = 2): string => {
+  const numericPrice = price ? parseFloat(price) : 0;
```
dmkirshon: Should this default to 0? This might cause issues as the receipt would show 0 when they get charged with an actual Ada value? Or maybe it just gets handled in the `PriceSummaryDialog` to an Error, contact support, or unavailable?

apps/studio/src/pages/home/uploadSong/PriceSummaryDialog.tsx
```diff
@@ -0,0 +1,240 @@
+import { FunctionComponent } from "react";
+import { useFormikContext } from "formik";
+import {
+  Dialog,
```
dmkirshon: Noticed you didn't use the elements>styled `Dialog.tsx`, is this another element we should transition out. Looks like the main difference is the styled background colors, which shows in between the delete account dialog and the price summary dialog.

dmkirshon (approved):
LGTM, just a couple comments for styling and edge case.

packages/utils/src/lib/string.ts
escobarjonatan: I modified it so it doesn't return the 0.
in case it gets a 0, it should return "0.00" which is correct. If not a number, it returns NaN and we can handle the logic at the consumer level.
In `PriceSummaryDialog ` i have added a fallback to `N/A` for now, we can update it if needed, i let product know

apps/studio/src/pages/home/uploadSong/PriceSummaryDialog.tsx
escobarjonatan: I didnt realize i was pulling it from mui. We should be pulling it from elements, updated it.

PR #445, opened 2023-12-08
Add isni and ipi fields to profile and default when uploading a song

escobarjonatan:
https://github.com/projectNEWM/newm-web/assets/11530120/137e78f3-7694-49f4-8f70-f5aee496079d


newmskywalker:
Task linked: [CU-86a1bbh1g Add Secondary Email, ISNI, and IPI fields to Profile page.](https://app.clickup.com/t/86a1bbh1g)

apps/studio/src/modules/song/thunks.ts
```diff
@@ -89,9 +89,9 @@ export const uploadSong = createAsyncThunk(
       let ipis;
 
       if (body.ipis) {
-        ipis = body.userIpi ? [...body.ipis, body.userIpi] : body.ipis;
```
escobarjonatan, line 92: with the introduction of the true user's ipi, i decided to rename this to ipi as it represents the ipi of the song, i believe this will make it less confusing, but if it's not lmk i'll revert it 🤷🏽‍♂️

dmkirshon: I think my original thought was that a song ipi will be the user's ipi and wanted to give that distinction from the other ipis. Also the one letter difference might cause some reading/updating confusions. I can agree in this context it would just be fine as just ipi.

apps/studio/src/pages/home/profile/Profile.tsx
```diff
@@ -275,6 +282,8 @@ const Profile: FunctionComponent = () => {
             { error: errors.firstName, element: firstNameRef.current },
             { error: errors.lastName, element: lastNameRef.current },
             { error: errors.companyName, element: companyNameRef.current },
+            { error: errors.isni, element: isniRef.current },
+            { error: errors.ipi, element: userIpiRef.current },
```
dmkirshon, line 286: Can remove `user` from ipi to follow suit with the refactor done in the other files.

dmkirshon (approved):
LGTM, just a few comments. 

Also, just wanted to confirm business logic:
- Does adding the `ipi` during the distribute & mint flow on the advanced details page then add it to the users profile?

apps/studio/src/modules/song/thunks.ts
escobarjonatan: An IPI will be the songs IPI. At the moment we are only allowing 1 to be submitted and it happens to be the users IPI, but at a later time I assume we will let them add others.

apps/studio/src/pages/home/profile/Profile.tsx
escobarjonatan: This is the users IPI therefore kept it as that. The reason why the field name is IPI is bc the BE expects it to be that

escobarjonatan:
> Does adding the ipi during the distribute & mint flow on the advanced details page then add it to the users profile?

Good question, at the moment it does not. That specific piece will be handled with a different ticket. There is a conversation in the ticket for this.

PR #459, opened 2023-12-12
Update sidebar NEWM icon and link to newm.io

dmkirshon:
Update:
- NEWM icon to NEWM logo from NEWM inverse logo on sidebar
- NEWM icon on sidebar to link to newm.io

### DIFF
--- a/apps/studio/src/pages/home/SideBar.tsx
+++ b/apps/studio/src/pages/home/SideBar.tsx
@@ -1,5 +1,5 @@
 import { FunctionComponent } from "react";
-import { Box, Drawer, IconButton, Stack, useTheme } from "@mui/material";
+import { Box, Drawer, IconButton, Link, Stack, useTheme } from "@mui/material";
 import { Typography } from "@newm-web/elements";
 import {
   SideBarHeader,
@@ -21,6 +21,7 @@ import {
 } from "@mui/icons-material";
 import {
   NEWM_CLICKUP_FORM_URL,
+  NEWM_IO_URL,
   NEWM_STUDIO_DISCORD_URL,
   NEWM_STUDIO_FAQ_URL,
 } from "../../common";
@@ -177,7 +178,9 @@ export const SideBar: FunctionComponent<SideBarProps> = (
         px={2.5}
         width="100%"
       >
-        <NEWMLogoSmInverse />
+        <Link href={NEWM_IO_URL} target="_blank" rel="noopener">
+          <NEWMLogoSmInverse />
+        </Link>
       </Box>
     </Box>
   );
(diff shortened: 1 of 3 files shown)


newmskywalker:
Task linked: [CU-86a0kw2v6 Update NEWM icon in side panel menu and hyperlink to newm.io website.](https://app.clickup.com/t/86a0kw2v6)

apps/studio/src/pages/home/SideBar.tsx
```diff
@@ -177,7 +178,9 @@ export const SideBar: FunctionComponent<SideBarProps> = (
         px={2.5}
         width="100%"
       >
-        <NEWMLogoSmInverse />
+        <Link href={NEWM_IO_URL} target="_blank" rel="noopener">
+          <NEWMLogo height="40" width="40" />
```
dmkirshon, line 192: There is an issue I can't identify the fix for, where the `linearGradient`, NEWM multi colored circle around the icon, isn't visible in the mobile view sidebar. I have tracked it down to the `Drawer` component and the `variant="temporary` attribute but haven't found anything leading to a solution online. I will add a notice on the task and create a separate low priority bug for it. We can also look at just using a png or other format for this small icon.

dmkirshon: Looked into this deeper and looks like it is an issue with the SVG ID being used within the `linearGradient` and how  the `Drawer` component must render SVGs causing a conflict. 

I will look into creating SVGs with a dynamic hash for the ID but if it's too much effort for this task I will sideline it to a bug.
https://stackoverflow.com/questions/14773940/conflicting-ids-in-two-different-svg-files-included-in-the-same-html-document

packages/assets/src/lib/images/NEWMLogo.tsx
```diff
@@ -8,55 +8,59 @@ interface NEWMLogoProps {
...
-      </linearGradient>
-      <clipPath id="clip0_3_842">
-        <rect width="150" height="150" fill="white" />
-      </clipPath>
-    </defs>
-  </svg>
-);
+}) => {
+  // create unique id for each instance of NEWMLogo to avoid id collisions
+  const id = useId();
```
dmkirshon, line 13: Added this as a temp fix for creating a unique ID for the `linearGradient` because of the `Drawer` component creating a duplicate of the sidebar within the DOM. Need to investigate further if a rule set for React SVG Components can be created to generate unique IDs on build.

apps/studio/src/pages/home/SideBar.tsx
dmkirshon: Resolved the issue with a temp fix in the next commit.

packages/assets/src/lib/images/NEWMLogo.tsx
dmkirshon: Removed the main issue of duplication for the `Drawer` component between responsive views by rendering the correct `Drawer` component based on the theme breakpoint. Leaving the `useId` as we might have other uses or locations of using the NEWM logo twice in the DOM. We can remove it once we have a more robust unique ID system for SVGs.

scandycuz (approved):

escobarjonatan (approved):

PR #470, opened 2023-12-20
Add prettier-eslint into monorepo

dmkirshon:
Add:
- Original ESLint and prettier rule sets from newm-artist-portal repo
   - [Original ESLint](https://github.com/projectNEWM/newm-web/blob/19dc616045fbf9f399a50beaabe2ca5689689ae1/.eslintrc.js)
      - Did not include `ignorePatterns`
   - [Original Prettier](https://github.com/projectNEWM/newm-web/blob/19dc616045fbf9f399a50beaabe2ca5689689ae1/.prettierrc.json)
- `"react/jsx-sort-props"` to sort Component props
- [sort-keys-fix](https://www.npmjs.com/package/eslint-plugin-sort-keys-fix) for sorting objects and other key pairs
   - standard sort-keys does not auto-fix
   - sort-keys only sorts a max of 10 at a time, requires multiple formats/saves to fix larger key sets
- [Typescript Interface key sorter](https://www.npmjs.com/package/eslint-plugin-typescript-sort-keys) 
- `"ignorePattern": "d=\"([\\s\\S]*?)\"` for  `"max-len"` rule to stop Asset SVG warning, in place for ignorePatterns at top level

Modified:
- `settings.json` to use prettier-eslint formatting

Notes: 

1. ESLint was erroring out with old config because `"plugin:storybook/recommended"` was missing from package.json, added back into dev-dependencies
2. `nx format <app-or-package>` command does not format the code in the same manner as `prettier-eslint`, avoid using the command to format all the code. Save on format will work in normal dev work flows. If monorepo formatting is needed, use the `npx prettier-eslint  --write "**/*.{jsx,js,ts,tsx}"` command


package.json
```diff
@@ -3,7 +3,9 @@
   "version": "0.0.0",
   "private": true,
   "license": "MIT",
-  "scripts": {},
+  "scripts": {
+    "format": "prettier-eslint --config .prettierrc.json --write \"**/*.{jsx,js,ts,tsx}\""
```
dmkirshon: This will format every single `jsx, js, ts, tsx` within newm-web, should we isolate or should this be fine

.vscode/settings.json
```diff
@@ -1,9 +1,15 @@
 {
-  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.defaultFormatter": "rvest.vs-code-prettier-eslint",
+  "editor.formatOnPaste": false, // required
+  "editor.formatOnType": false, // required
+  "editor.formatOnSave": true, // optional
+  "editor.formatOnSaveMode": "file", // required to format on save
+  "files.autoSave": "onFocusChange", // optional but recommended
+  "vs-code-prettier-eslint.prettierLast": false, // set as "true" to run 'prettier' last not first
```
dmkirshon, line 8: I added these from my original workspace settings for `prettier-eslint` as I believe they were recommended to add in, let me know and we can modify as needed

.eslintrc.json
```diff
@@ -24,7 +34,97 @@
...
+            "ignoreDeclarationSort": true,
+            "ignoreMemberSort": false
+          }
+        ],
+        "sort-keys-fix/sort-keys-fix": [
+          "error",
+          "asc",
+          { "caseSensitive": true, "natural": false }
+        ]
+      },
```
dmkirshon, line 151: `sort-keys` but with a fix to handle auto-fix on 

Note: There is a small quirk for this, I believe it only sorts up to 10 key sets at a time, so requires the format script to be run a few times for larger out of order sets, can be mostly mitigated with running the format script multiple times initially to make sure all sorting is complete. Errors will always show in the file if sorting hasn't been completed on first save so Linting still will save us from any deviations.

.eslintrc.json
```diff
@@ -24,7 +34,97 @@
...
+        "react/jsx-sort-props": [
+          2,
+          {
+            "callbacksLast": true,
+            "shorthandFirst": false,
+            "shorthandLast": true,
+            "ignoreCase": true,
+            "noSortAlphabetically": false
+          }
+        ],
```
dmkirshon, line 135: Prop sorter for components, can modify options as preferred

.eslintrc.json
```diff
@@ -24,7 +34,97 @@
...
+        "react/prop-types": "warn",
+        "semi": "warn",
+        "sort-imports": [
+          "warn",
+          {
+            "ignoreCase": false,
+            "ignoreDeclarationSort": true,
+            "ignoreMemberSort": false
+          }
+        ],
```
dmkirshon, line 145: I tested out another import sorter: `“simple-import-sort”` as noted in https://medium.com/aeturnuminc/eslint-configuration-for-react-formatters-and-rules-8ea0a29dd97

I preferred the layout and spacing of the `“simple-import-sort”`, I will see if the same options can be applied for our current `"sort-imports"` otherwise might recommend moving to the referenced sorter. 

I believe the only difference is the `allowSeparatedGroups` option to be true which I would recommend turning on, gives just a bit of spacing to see the import source differences.

.eslintrc.json
```diff
@@ -1,7 +1,17 @@
...
+  "extends": [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:react/recommended",
+    "plugin:jsx-a11y/recommended",
+    "plugin:react-hooks/recommended",
+    "plugin:jest/recommended",
+    "plugin:testing-library/react",
+    "plugin:react/jsx-runtime"
+  ],
```
dmkirshon: Removed `"plugin:storybook/recommended"` while migrating from the original `.eslintrc.json` since the package.json dependency was missing and causing the formatting to error out, can add back in if deemed necessary

.eslintrc.json
```diff
@@ -1,7 +1,17 @@
...
   "root": true,
   "ignorePatterns": ["**/*"],
-  "plugins": ["@nx"],
+  "plugins": ["@nx", "@typescript-eslint", "import", "sort-keys-fix"],
```
dmkirshon: I did not bring over the original `.eslintrc.json` settings for `ignorePatterns`, `parser`, `parserOptions`, and `env`, let me know if they are required but they did not affect my initial formatting test
Original Config: https://github.com/projectNEWM/newm-web/blob/19dc616045fbf9f399a50beaabe2ca5689689ae1/.eslintrc.js

.eslintrc.json
```diff
@@ -24,7 +34,97 @@
...
+              "internal"
+            ]
+          }
+        ],
+        "jsx-quotes": ["warn", "prefer-double"],
+        "max-len": [
+          "warn",
+          {
+            "code": 120,
+            "ignorePattern": "d=\"([\\s\\S]*?)\""
```
dmkirshon, line 98: This was put in place instead of setting the `ignorePattern` at the top level for `Assets` like the original `.eslintrc.json`, it skips SVG path warnings. We could also place `"ignoreStrings": true` but this would affect copy text as well.

package.json
scandycuz: I would put this in the Makefile instead. That is what Nx recommends, as it seems to prefer having scripts for the root of the repo there and `project.json` files for the individual apps/packages. That way all of the scripts are in the same places.

scandycuz (changes requested):
LGTM, but had a comment about where the formattings script should go.

package.json
dmkirshon: I was able to add the app/package specific commands to the `project.json` files, I also added the documentation to our readme to run format across any or all apps/packages.

- Single App or Package: `nx run <app-or-package>:format`
- All Apps and Packages: `nx run-many -t format`

I attempted but was unable to get anything configured in the Makefile. I'm unsure of what to actually include or how to implement the formatting through the Makefile. I did find that nx does have docs on implementing root level NPM scripts: https://nx.dev/recipes/running-tasks/root-level-scripts, if we want to do a format for the entire `newm-web`, but with the above format commands I don't think it's necessary. Let me know your thoughts.

.vscode/settings.json
escobarjonatan: I think the following can be left out as i feel it is a little opinionated(i have these exact same preferences/values locally). I've ran into cases where i need to turn them on/off as im working, having these here would conflict with user's local config. I can always modify them here i guess 🤷🏽‍♂️
```
 "editor.formatOnPaste": false, // required
  "editor.formatOnType": false, // required
  "editor.formatOnSave": true, // optional
  "editor.formatOnSaveMode": "file", // required to format on save
  "files.autoSave": "onFocusChange", // optional but recommended
```

escobarjonatan (approved):

package.json
scandycuz: The `project.json` approach works. I wouldn't add anything to the README though. The command to run the format script is the same as running other scripts in the `project.json` files, which is covered by the "Running tasks" section, and it will probably only ever be run once. The storybook command is only included in the README because it's something that someone looking to contribute to or learn about the repo may want to run.

dmkirshon: Got it, will roll back the README change

.vscode/settings.json
dmkirshon: I was trying to remember where these settings came from. These are actually recommended by the `Prettier ESLint` extension with the exact comments, I would rather keep them if they are recommended for the configuration of the project. Let me know if you feel otherwise, I have these exact settings from the artist-portal and never ran into issues...

.eslintrc.json
```diff
@@ -1,7 +1,47 @@
...
+  "extends": [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:react/recommended",
+    "plugin:jsx-a11y/recommended",
+    "plugin:react-hooks/recommended",
+    "plugin:jest/recommended",
+    "plugin:testing-library/react",
+    "plugin:react/jsx-runtime",
+    "plugin:typescript-sort-keys/recommended",
```
dmkirshon, line 23: Added additional sorter for TypeScript Interfaces

.vscode/settings.json
scandycuz: I prefer to have them in the repo, that way all developers are using the same settings. If there is a specific rule we'd prefer not to use as a team, then we can remove it.

scandycuz (approved):
Working for me locally 👍

PR #475, opened 2024-01-16
Add a disabled delete button on the view details page and misc updates

escobarjonatan:
Added a disabled delete button to the view song details page
Removed duplicated imports

### DIFF
--- a/apps/studio/src/components/stories/form.stories.tsx
+++ b/apps/studio/src/components/stories/form.stories.tsx
@@ -1,9 +1,8 @@
 import { Box } from "@mui/material";
 import { Form, Formik, FormikHelpers } from "formik";
-import { Button, Typography } from "@newm-web/elements";
+import { Button, TextInputField, Typography } from "@newm-web/elements";
 import * as Yup from "yup";
 import theme from "@newm-web/theme";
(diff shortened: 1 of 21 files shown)


apps/studio/src/components/stories/form.stories.tsx
dmkirshon, line 1: Understand this was just an edit to handle the imports but I'm getting some errors when this file is open. I'm assuming it's because the storybook structure moved to the `elements` package and configs/references aren't setup correctly within the `studio` app. Do we still want to have this story or can we clean it up as well?

Imports:
`Cannot find module '@newm-web/theme' or its corresponding type declarations.ts(2307)`
`Cannot find module '@newm-web/elements' or its corresponding type declarations.ts(2307)`
Components:
`'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.ts(2686)`

dmkirshon (approved):
LGTM, just one non-blocking issue/question for the `form.stories.tsx` file.

apps/studio/src/components/stories/form.stories.tsx
escobarjonatan: Moved this story to the `elements` stories as this is where its dependencies are. I think it's still useful no need to delete it for now.

PR #478, opened 2024-01-19
fix: Alt copy for songs uploaded but no collabs

dmkirshon:
Fix:
- Show alternative copy for the Upload a Song navigation button within the no collabs component (`NoOwnersYet.tsx`), on the collaborator page, when songs are already uploaded to the Artists account but no collabs are attached to any of the songs

https://projectnewm.atlassian.net/browse/STUD-59

### DIFF
--- a/apps/studio/src/pages/home/owners/OwnersTable.tsx
+++ b/apps/studio/src/pages/home/owners/OwnersTable.tsx
@@ -20,7 +20,10 @@ import {
 import AccountCircleIcon from "@mui/icons-material/AccountCircle";
 import NoOwnersYet from "./NoOwnersYet";
 import { history } from "../../../common/history";
-import { useGetCollaboratorsQuery } from "../../../modules/song";
+import {
+  useGetCollaboratorsQuery,
+  useGetSongCountQuery,
+} from "../../../modules/song";
 
 interface OwnersTableProps {
   query: string;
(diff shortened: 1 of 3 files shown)


apps/studio/src/pages/home/owners/OwnersTable.tsx
```diff
@@ -110,7 +120,7 @@ export default function OwnersTable({
   }
 
   if (isSuccess && collaboratorsData?.length === 0 && !query) {
```
dmkirshon, line 124: Just wanted to confirm the default behavior that currently exists: if getting the collaborator data is not a success then a table with default copy appears. If you agree something else should render on network/data query failure I can create a ticket to investigate further.

scandycuz (approved):
LGTM. One syntax suggestion I'd make is to pass in a `hasSongs` boolean prop to the `NoOwnersYet` component, instead of the `totalCountOfSongs` number prop, since the component is only concerned about whether the song count is zero or not.

dmkirshon:
> LGTM. One syntax suggestion I'd make is to pass in a `hasSongs` boolean prop to the `NoOwnersYet` component, instead of the `totalCountOfSongs` number prop, since the component is only concerned about whether the song count is zero or not.

@scandycuz, refactored with hasSongsUploaded for more clarity on prop being used for the button copy conditional.

PR #487, opened 2024-01-23
fix: Validate wallet and profile only in mint form

dmkirshon:
fix:
- Update conditionals to only show wallet and profile verification alerts to user while they're completing the Distribute & Minting form.
- Update next step button conditional to only be disabled when a wallet isn't connected or profile verfication hasn't been completed when the user is completing the Distribute & Mint form 

Note: Currently the business logic doesn't make the Distribute & Mint form visible within the `MINTING` tab. This is done via the `edit` page not the `view-details` page. I chose to keep the logic as is since the JIRA request was to remove the alerts from being visible when Minting had already been initiated. A ticket will be created to address removing the Distribute & Mint form logic within `MintSong.tsx` or otherwise consolidating the `edit` and `view-details` pages into one page.

https://projectnewm.atlassian.net/browse/STUD-50

### DIFF
--- a/apps/studio/src/pages/home/library/MintSong.tsx
+++ b/apps/studio/src/pages/home/library/MintSong.tsx
@@ -281,6 +281,12 @@ const MintSong = () => {
             ? dirty
             : values.isMinting;
 
+          // if minting has been toggled but mint hasn't been initiated
+          const isMintingFormVisible = !isMintingInitiated && values.isMinting;
+
+          const isStepOneButtonDisabled =
+            isMintingFormVisible && (!isVerified || !wallet);
+
           const handleChangeOwners = (values: ReadonlyArray<Owner>) => {
             setFieldValue("owners", values);
           };
@@ -347,7 +353,7 @@ const MintSong = () => {
                       </Box>
                     ) }
 
-                    { !isMintingInitiated && values.isMinting && !isVerified && (
+                    { isMintingFormVisible && !isVerified && (
                       <Alert
                         action={
                           <Button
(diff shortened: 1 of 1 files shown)


apps/studio/src/pages/home/library/MintSong.tsx
```diff
@@ -424,7 +424,9 @@ const MintSong = () => {
 
                     { isStepOneButtonVisible && (
                       <Button
-                        disabled={ !isVerified || !wallet }
+                        disabled={
+                          !isMintingInitiated && (!isVerified || !wallet)
```
dmkirshon: Was deciding if this conditional is too convoluted with all the `!`, was thinking of these alternatives:

1. Create a new reference `const isMintingFormVisible = !isMintingInitiated && isMinting`. Then replace it here and for the alerts above. The updated line would be: 
`disabled={ isMintingFormVisible && (!isVerified || !wallet)`

2.  This one might convolute even more but reversing the entire conditional and putting the `!` outside to basically state this is not disabled when this condition is met. 
`disabled={ !(isMintingInitiated || (isVerified && wallet)`

escobarjonatan: I like option 1 plus adding a variable so all conditional logic is added in there and this just reads:

`const shouldDisableNextAction = (!isMintingInitiated && isMinting) && (!isVerified || !wallet)`
`disabled={shouldDisableNextAction}` 

or something along these lines.

escobarjonatan (approved):

PR #488, opened 2024-01-23
Update logic to min scheduled release date validation

escobarjonatan:


apps/studio/src/common/formUtils.ts
```diff
@@ -219,7 +219,7 @@ export const commonYupValidation = {
     return Yup.date()
       .required("This field is required")
       .min(
-        new Date(minReleaseDate.getTime() - 24 * 60 * 60 * 1000),
```
escobarjonatan, line 222: Not sure where this subtraction of date came from, but was causing issues with validation of the scheduled release date

dmkirshon: So I originally added it back on:
https://github.com/projectNEWM/newm-web/commit/9169569675e99f39d47bd69d4d0d3ab406ee7a95

This is because you get an issue like the one below in the image. The validation copy doesn't line up with the date and what is available to choose from the calendar, I believe it was due to either the Date Input Field attribute field `min` or the Yup validation `min` not being inclusive. Which leads to the incorrect copy/validation issue. My bad on the comment above the validation, it could have been more robust on explaining that situation. Need to find a way to marry the two so it aligns without causing the original bug.

image

dmkirshon (changes requested):
Requires a solution to fix another issue that was being prevented by the one day off code.

apps/studio/src/common/formUtils.ts
dmkirshon: Maybe a custom Yup validation `test` using a conditional on the dates instead of using `min` to validate the date?

dmkirshon: Maybe a custom Yup validation `test` using a conditional on the dates instead of using the Yup `min` validation to validate the date? That way we could custom test the Yup validation to match the inclusive conditional of the Date Input `min` attribute.

dmkirshon: Identified as a timezone issue:

Ran console.logs for the code below, looks like the entered release date is using the local time while the minReleaseDate is using UTC/GMT. Since it's a 2 hour time difference ahead of UTC/GMT on my side the comparison fails for the day of.
1707256800000 (My Time)  => 1707264000000 (Min Release Date)  which returns false.

```
.test("is-valid-releaseDate", "Invalid release date", (value) => {
        const enteredReleaseDate = new Date(value);
        return enteredReleaseDate.getTime() >= minReleaseDate.getTime();
      });
```

apps/studio/src/common/formUtils.ts
```diff
@@ -209,17 +209,25 @@ export const commonYupValidation = {
...
-    // If releaseDate is provided, use it.
-    // Otherwise, use the minimum time for EVEARA to distribute from now.
+    // This is necessary to avoid time zone discrepancies,
+    // as Date object uses the local time zone by default.
+    const parseDateAsUTC = (dateString: string) => {
+      const [year, month, day] = dateString.split("-");
+      return new Date(
+        Date.UTC(parseInt(year), parseInt(month) - 1, parseInt(day))
+      );
+    };
```
dmkirshon: I don't think the `parseDateAsUtc` is needed at all. The function is doing it's job to convert to UTC but the conversion doesn't really matter in the `min` comparison below because ISO already converts the release date `Date` object to UTC and `.split("T")[0]` provides it in the same format as in the Field Value. It just matters that the Dates are formatted in the same YEAR-MONTH-DAY format at that point of comparison. 

The main issue before was that we were comparing Unix values, including hour of the day in the local area, which would not be valid. Really the big difference is now the comparison is less rigid by only using dates.

```
releaseDateFieldValue -> "2024-02-12"

new Date(minReleaseDate).toISOString().split("T")[0]  -> '2024-02-12'
```

You can just remove the `parseDateAsUTC` and change the `minReleaseDate` ternary back to `new Date(releaseDate)` and I believe this will all still work.

apps/studio/src/common/formUtils.ts
```diff
@@ -209,17 +209,25 @@ export const commonYupValidation = {
   password: Yup.string().required("Password is required"),
   publicationDate: Yup.date().max(new Date(), "Cannot be a future date"),
   releaseDate: (releaseDate: string | undefined) => {
```
dmkirshon, line 211: Maybe we should rename the `releaseDate` argument to something that actually references that it's the provided minReleaseDate from the backend endpoint. Right now it kind of looks like the field value, which it's not. Maybe `minReleaseDateString`?

escobarjonatan: I dont necessarily agree, we would just be adding "string" to a string in different context. will keep as is

dmkirshon (approved):
LGTM

PR #493, opened 2024-01-29
MRKT-28: implements blank home page with logo

scandycuz:
Adds blank Marketplace home page with header and logo.

### Also

- Refactors asset package to move icons into icons directory

Screen Shot 2024-01-28 at 11 37 31 PM


packages/assets/src/lib/images/NEWMMarketplaceLogo.tsx
dmkirshon, line 1: If we follow the same pattern of using MUI icons it might be worth going through the MUI svg icon creation pattern. This one is pretty custom so can fall out of the pattern.

```
import { createSvgIcon } from "@mui/material";

const NEWMMarketplaceLogo = createSvgIcon(
```

packages/assets/src/index.ts
```diff
@@ -16,6 +16,7 @@ import InfoCircleLine from "./lib/images/InfoCircleLine";
 import InstagramLogo from "./lib/images/InstagramLogo";
 import NEWMLogo from "./lib/images/NEWMLogo";
 import NEWMLogoSmInverse from "./lib/images/NEWMLogoSmInverse";
+import NEWMMarketplaceLogo from "./lib/images/NewmMarketplaceLogo";
```
dmkirshon: I'm getting a `Module not found` compile error due to the lower case `Newm` instead of `./lib/images/NEWMMarketplaceLogo`

dmkirshon (approved):
Just one fix needed that led to a compile error for me, and a potential pattern change for the SVG. Otherwise LGTM!

Also noticed that the NEWM Studio firebase action tried to run and failed, we will need to customize it to handle the Marketplace PRs.

.github/workflows/firebase-hosting-pull-request-newm-studio.yml
```diff
@@ -4,7 +4,6 @@ on:
     paths:
       - .github/workflows/firebase-hosting-pull-request-newm-studio.yml
       - apps/studio/**
-      - packages/**
```
scandycuz, line 7: Disables PR preview builds if only packages are updated. Could be worth adding a separate Github workflow to deploy to both a NEWM Studio and NEWM Marketplace preview url if a package is updated.

scandycuz: Also, currently in the process of getting staging and preview urls for NEWM Marketplace setup. The Github workflows for deploying to them will be implemented in a separate PR.

dmkirshon (approved):
LGTM

PR #496, opened 2024-02-01
Update social create account and login options

dmkirshon:
Remove:
- Sign in with Facebook
- Sign in with LinkedIn

Add:
- Sign in with Apple

Note: The Sign In with Apple pop up comes up but we run into a similar localhost issue as we did for Sign in with LinkedIn. I attempted to be sneaky by manually setting the redirectUri, getting the valid pop up, logged in and verfied through device, but couldn't get past an error: "Your request could not be completed because of an error. Please try again later." This is probably due to the redirect issue and the PR might need to be accepted onto Master or have the Server team temporarly add the PR hosted address to fully test the NEWM API endpoint.

https://projectnewm.atlassian.net/browse/STUD-58


.github/workflows/firebase-hosting-deploy-published-release.yml
```diff
@@ -11,9 +11,8 @@ jobs:
       - run: npm install
       - run: npx nx build-with-sentry studio
         env:
-          VITE_FACEBOOK_CLIENT_ID: ${{ secrets.FACEBOOK_CLIENT_ID }}
+          VITE_APPLE_CLIENT_ID: ${{ secrets.GOOGLE_CLIENT_ID }}
```
dmkirshon: Need to change this to `APPLE_CLIENT_ID`

apps/studio/src/components/AppleLogin.tsx
```diff
@@ -0,0 +1,65 @@
...
+interface Props {
+  readonly children?: ReactNode;
+}
+
+const AppleLogin: FunctionComponent<Props> = ({ children }) => {
+  useScript(appleAuthHelpers.APPLE_SCRIPT_SRC);
+  const dispatch = useAppDispatch();
+  const [logIn] = useAppleLoginThunk();
+
+  const redirectUri = `${window.location.origin}/apple`;
```
dmkirshon: I think this should be `window.location.href` instead of apple so it matches the appropriate page where the button/function is being called? We have the two sign up paths already enabled through the Apple dev backend as well. This format was used in the previous linkedIn file but I think we were using a `LinkedInCallback` route with a `/linkedin` path.

apps/studio/src/components/AppleLogin.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const handleLogin = () =>
+    appleAuthHelpers.signIn({
```
dmkirshon, line 30: This actually returns a Promise but I saw it done like this for the other logins, should I use an async/await instead? I don't think there is any issues currently as it waits for the onSuccess callback anyway.

.github/workflows/firebase-hosting-deploy-published-release.yml
dmkirshon: Fixed.

apps/studio/src/components/AppleLogin.tsx
dmkirshon: Updated to `window.location.href` to match the location of the NEWM login and sign-up pages where the apple sign will be located.

apps/studio/src/components/AppleLogin.tsx
scandycuz: Should be fine without async/await. The function's still returning a promise, async/await is only really necessary when you need to wait for a promise to complete in the code.

apps/studio/src/components/AppleLogin.tsx
```diff
@@ -24,7 +24,7 @@ const AppleLogin: FunctionComponent<Props> = ({ children }) => {
   const dispatch = useAppDispatch();
   const [logIn] = useAppleLoginThunk();
 
-  const redirectUri = `${window.location.origin}/apple`;
```
scandycuz: I'd prefer to be explicit about what the redirect uri is, since it needs to match what we have on the Apple side. If this component was moved to a different route, the implementation would no longer work. 

I'd prefer either `${window.location.origin}/login` or `window.location.origin`.

scandycuz (approved):

PR #502, opened 2024-02-06
MRKT-29: connect wallet

scandycuz:
Screen Shot 2024-02-05 at 10 57 43 PM

Screen Shot 2024-02-05 at 10 58 03 PM

---

Implements connect and disconnect wallet functionality.

### Also

- Refactors env vars. It will be necessary to add additional variables to the `.env` file (as outlined in `.env.template`) for use in Next.js.


apps/marketplace/src/store.ts
```diff
@@ -0,0 +1,24 @@
...
+      : baseMiddleware;
+  },
+  reducer,
+});
+
+export type AppDispatch = typeof store.dispatch;
+
+export type RootState = ReturnType<typeof store.getState>;
+
+export default store;
```
scandycuz, line 24: I didn't end up using Redux, but I figured I'd leave some of the code, since we'll need it eventually.

apps/studio/public/index.html
scandycuz, line 1: It appears that this `index.html` file wasn't actually being used.

apps/marketplace/src/app/layout.tsx
```diff
@@ -13,12 +15,33 @@ interface RootLayoutProps {
...
+          href={
+            "https://fonts.googleapis.com/css2?" +
+            "family=DM+Serif+Text:ital@0;1&" +
+            "family=Inter:wght@400;500;600;700;800&" +
+            "family=Raleway:wght@400;500;600;700;800&" +
+            "display=swap"
+          }
+          rel="stylesheet"
+        />
+      </head>
```
escobarjonatan, line 35: I think we can use the Head component from nextjs here: https://nextjs.org/docs/pages/api-reference/components/head

escobarjonatan (approved):

apps/marketplace/src/app/layout.tsx
scandycuz: The component looks to be for the legacy "pages" router, as apposed to the new  "app" router, which is recommended for new projects. I actually tested out that component, but it didn't add the tags to the html file. The app router looks to suggest just using normal html tags in the layout components.

scandycuz: The app router also provides a [meta-data API](https://nextjs.org/docs/app/building-your-application/optimizing/metadata) to add meta tags like the page title, but stylesheet links are not currently supported.

scandycuz: There's also an [API for Google fonts](https://nextjs.org/docs/app/building-your-application/optimizing/fonts), but it doesn't work with how we specify font-family styles in the theme.

PR #504, opened 2024-02-07
MRKT-39: deploy marketplace to staging

scandycuz:
### Updates

- Deploys NEWM Marketplace to a preview url when a PR is created.
- Deploys NEWM Marketplace to staging (fan.square.newm.io) when a PR is merged into master.

### Notes

- When a package is changed, both NEWM Studio and NEWM Marketplace will be deployed.

### DIFF
--- a/.github/workflows/firebase-hosting-merge-deploy-newm-marketplace.yml
+++ b/.github/workflows/firebase-hosting-merge-deploy-newm-marketplace.yml
@@ -1,11 +1,11 @@
-name: Deploy NEWM Studio to Firebase Hosting on merge
+name: Deploy NEWM Marketplace to Firebase Hosting on merge
 on:
   push:
     branches:
       - master
     paths:
-      - .github/workflows/firebase-hosting-merge-deploy-newm-studio.yml
-      - apps/studio/**
+      - .github/workflows/firebase-hosting-merge-deploy-newm-marketplace.yml
+      - apps/marketplace/**
       - packages/**
 jobs:
   build_and_deploy:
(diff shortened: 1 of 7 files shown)


.github/workflows/firebase-hosting-merge-deploy-newm-marketplace.yml
```diff
@@ -0,0 +1,43 @@
...
+          NEXT_PUBLIC_ENV: staging
+
+      - uses: FirebaseExtended/action-hosting-deploy@v0
+        id: deploy
+        with:
+          repoToken: "${{ secrets.GITHUB_TOKEN }}"
+          firebaseServiceAccount: "${{ secrets.FIREBASE_SERVICE_ACCOUNT_NEWM_ARTIST_PORTAL }}"
+          channelId: live
+          projectId: "${{ secrets.FIREBASE_PROJECT_ID_NEWM_ARTIST_PORTAL }}"
+          target: marketplace
```
dmkirshon, line 31: Just wanted to confirm, the `NEWM_ARTIST_PORTAL` references are due to the two sites sharing the same Firebase project?

dmkirshon (approved):
LGTM, just one question

.github/workflows/firebase-hosting-merge-deploy-newm-marketplace.yml
scandycuz: Yeah, there are only two Firebase projects, one for the staging environment, and one for the production environment. Each project has multiple "targets" which is where each app is deployed. We should probably update the secret names to make it more obvious that they are for all apps in an environment, but that can be done as part of a separate PR.

PR #505, opened 2024-02-07
Update profile outlet copy and add warning message

escobarjonatan:
feat: STUD-79, STUD-107, STUD-108. Update profile outlet copy and add warning message

Profile section updates:
image

Profile outlet warning message:
image

### DIFF
--- a/apps/studio/src/pages/home/profile/Profile.tsx
+++ b/apps/studio/src/pages/home/profile/Profile.tsx
@@ -1,4 +1,5 @@
-import { FunctionComponent, useRef } from "react";
+import { FunctionComponent, useEffect, useRef } from "react";
+import { useLocation } from "react-router-dom";
 import {
   Box,
   Container,
@@ -55,6 +56,7 @@ const { Unverified, Pending, Verified } = VerificationStatus;
 
 const Profile: FunctionComponent = () => {
   const dispatch = useAppDispatch();
+  const { state } = useLocation();
 
   const companyNameRef = useRef<HTMLInputElement>(null);
   const emailRef = useRef<HTMLInputElement>(null);
@@ -63,6 +65,7 @@ const Profile: FunctionComponent = () => {
   const roleRef = useRef<HTMLDivElement>(null);
   const isniRef = useRef<HTMLInputElement>(null);
   const userIpiRef = useRef<HTMLInputElement>(null);
+  const outletProfilesRef = useRef<HTMLDivElement>(null);
 
   const windowWidth = useWindowDimensions()?.width;
   const { data: roles = [] } = useGetRolesQuery();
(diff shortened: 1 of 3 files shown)


apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
```diff
@@ -124,6 +128,18 @@ const BasicSongDetails: FunctionComponent<BasicDonDetailsProps> = ({
...
+    navigate("/home/profile");
+
+    setTimeout(() => {
+      const element = document.getElementById("outlet-profiles");
+
+      if (element) {
+        element.scrollIntoView({ behavior: "smooth" });
+      }
+    }, 0);
+  };
```
escobarjonatan: I hadn't ran into a case to use the `useNavigate` to navigate to a part of the page before but the internet seemed to like this workaround. Let me know if you all have ran into the need to navigate to a specific part of the page with the hook and have a different solution i am open to it

apps/studio/src/pages/home/profile/Profile.tsx
```diff
@@ -461,13 +462,26 @@ const Profile: FunctionComponent = () => {
...
                         </Typography>
                         <Typography variant="subtitle2">
-                          Connecting your outlet profile ensures that your music
-                          will be associated with your designated artist page on
-                          each platform following distribution.
+                          Linking your artist profile ensures your music will be
+                          linked to your designated artist page on each platform
+                          post-distribution. For details on finding your outlet
+                          profile and other requirements,
+                          <Link
```
scandycuz, line 485: Minor thing, but there looks to be a space missing before the link text.

apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
scandycuz: It would be ideal if it scrolled to a bit above the element, as apposed to it being right at the top of the screen. May not be worth the effort to implement though, and it could always be tweaked later.

scandycuz: An alternative approach could be to pass a route state when navigating, e.g. something like `{ shouldScrollToOutlets: true }`, and then the profile page could scroll to the location using a ref if it's present. Might be a little more "Reacty", not sure it's a better solution though.

scandycuz (commented):
LGTM, I think the only real change that would be good to make is adding a space before the "refer to our FAQ" link.

dmkirshon (approved):

dmkirshon (approved):
LGTM, just needs Trevor's comments addressed.

apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
escobarjonatan: Thanks for the suggestion, i went with the useLocation approach you suggested. Looks cleaner 🤷🏽‍♂️

apps/studio/src/pages/home/profile/Profile.tsx
escobarjonatan: Those characters were too small to notice lol

PR #506, opened 2024-02-07
Modify the scroll to top and message display logic

escobarjonatan:


apps/studio/src/pages/home/profile/Profile.tsx
```diff
@@ -236,7 +236,6 @@ const Profile: FunctionComponent = () => {
     if (state?.scrollToOutlets && outletProfilesRef.current) {
       outletProfilesRef.current.scrollIntoView({
         behavior: "smooth",
-        block: "center",
```
escobarjonatan, line 239: Denny wants it to scroll to the top and not the center

apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
```diff
@@ -79,7 +79,7 @@ const BasicSongDetails: FunctionComponent<BasicDonDetailsProps> = ({
   const { data: moodOptions = [] } = useGetMoodsQuery();
   const { data: languages = [] } = useGetLanguagesQuery();
   const { songId } = useParams<"songId">() as SongRouteParams;
-  const shouldShowOutletsWarning = !appleMusicProfile && !spotifyProfile;
```
escobarjonatan, line 82: This needs to be showed if any is missing not if both are missing only, added the || instead of the && to satisfy that

apps/studio/src/pages/home/profile/Profile.tsx
scandycuz: Would be great if there could be an inch of spacing or so above section after it scrolls, but I think that is a bit of work to implement.

scandycuz (approved):
LGTM

apps/studio/src/pages/home/profile/Profile.tsx
escobarjonatan: I mentioned that we were doing the scroll to center in error messages but he suggested we did this bc they're two different scenarios.... potato potatoe.

PR #511, opened 2024-02-08
MRKT-20: Footer

dmkirshon:
Screenshot from 2024-02-08 16-21-58
Screenshot from 2024-02-08 16-22-06

Add:
- Footer to Marketplace
- New externalLinks typography variant to theme
- Initial external link constants for NEWM Marketplace

https://projectnewm.atlassian.net/browse/MRKT-20


apps/marketplace/src/common/constants.ts
dmkirshon, line 1: Should this be consolidated into a `constants.ts` file under the `utils` package for Studio and Marketplace?

apps/marketplace/src/app/layout.tsx
```diff
@@ -34,12 +35,17 @@ const RootLayout: FunctionComponent<RootLayoutProps> = ({ children }) => {
         />
       </head>
 
-      <body style={ { backgroundColor: theme.colors.black } }>
+      <body
+        style={ {
+          backgroundColor: theme.colors.black,
```
dmkirshon, line 40: @scandycuz Is this just a formatting glitch between our two setups?

apps/marketplace/src/global.css
```diff
@@ -8,11 +8,14 @@ html {
   scroll-behavior: smooth;
 }
 body {
+  color: #ffffff;
+  display: flex;
+  flex-direction: column;
   font-family: inherit;
+  justify-content: space-between;
```
dmkirshon, line 15: Added these to line up the footer, but will probably need to be updated once page content is displayed.

packages/assets/src/lib/icons/XLogo.tsx
dmkirshon, line 1: Added a XLogo but didn't remove the original bird TwitterLogo as it's still being used in Studio. Left a comment for Design and Product to update in Studio Figma designs.

packages/theme/src/lib/theme.ts
```diff
@@ -260,30 +263,25 @@ const theme = createTheme({
...
-
     // custom font theme styles
+    // eslint-disable-next-line sort-keys-fix/sort-keys-fix
     emphasized: {
       fontFamily: "DM Serif Text",
       fontStyle: "italic",
       fontWeight: 400,
     },
 
+    externalLinks: {
```
dmkirshon: This is something I wanted to get on the same page with the design side so we are using a consistent typography across these type of links and designs. We can utilize/update this over in Studio as well when the opportunities come up.

apps/marketplace/src/common/constants.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * NEWM External Links and Support
+ */
+export const NEWM_DISCORD_URL = "https://discord.com/invite/z4sYqxqgvQ";
+export const NEWM_IO_URL = "https://newm.io/";
+export const NEWM_MARKETPLACE_FAQ_URL = "";
+export const NEWM_MARKETPLACE_SUPPORT_URL = "";
+export const NEWM_MARKETPLACE_TERMS_OF_SERVICE_URL = "";
```
dmkirshon, line 8: There is another tasking to fill these URLs in once they are addressed on Product/Legal side.

https://projectnewm.atlassian.net/browse/MRKT-37

apps/marketplace/src/components/footer/Footer.tsx
```diff
@@ -0,0 +1,93 @@
...
+      >
+        <Link href={ NEWM_IO_URL } rel="noopener" target="_blank">
+          <NEWMLogo height="40" width="40" />
+        </Link>
+        <Typography
+          component="div"
+          sx={ {
+            alignItems: "center",
+            display: "flex",
+            flexDirection: ["column", "row"],
```
dmkirshon, line 48: I made the Marketplace links a column as they didn't seem to center well, I thought they threw off the look. This can be updated as we do a more thorough Mobile responsive review with actual content being displayed.

apps/marketplace/src/common/constants.ts
escobarjonatan: I think that we can combine the ones that apply for both but keep the ones specific to each in their own app? I could see it becoming huge when we have more than these 2 apps 👀

We could even prefix global ones with `GLOBAL_`

apps/marketplace/src/components/footer/Footer.tsx
```diff
@@ -0,0 +1,93 @@
...
+  NEWM_MARKETPLACE_TERMS_OF_SERVICE_URL,
+  NEWM_TELEGRAM_URL,
+  NEWM_X_URL,
+} from "../../common/constants";
+
+const Footer: FunctionComponent = () => {
+  const theme = useTheme();
+
+  return (
+    <Grid
```
escobarjonatan, line 20: Semantically this can be a footer element. you can add `component="footer"` to this `Grid` for that.

escobarjonatan (commented):
minor comments, other than that this looks good

escobarjonatan (approved):

apps/marketplace/src/common/constants.ts
scandycuz: I prefer the approach of having code specific to certain apps in that app, unless there is a specific use case for why having them in a shared package would be beneficial.

scandycuz: I could see potentially having them in a shared file, but nesting them in an object specific to each app, e.g. `studio.EXAMPLE`.

packages/theme/src/lib/theme.ts
scandycuz: I might recommend making a custom `Link` component with the specified styling, as apposed to updating the theme. Same result, but a bit more modular. I could see the theme becoming a sort of large global stylesheet if we add too much to it.

scandycuz (approved):

apps/marketplace/src/app/layout.tsx
scandycuz: Did you already have it on multiple lines? It looks like the formatter won't enforce it being on a single line.

dmkirshon: No it was on a single line and the comma dangle was added with format-on-save, might have just been related to the other Format issues we've been seeing.

https://github.com/projectNEWM/newm-web/blob/dc189314b30db8171824633b0f24d38fe390bbdc/.eslintrc.json#L69-L70

apps/marketplace/src/common/constants.ts
dmkirshon: I'm going to leave it as is right now but as we build out the market/wallet features on Studio I believe we will get to a place where the Marketplace links might be required there as well. Would make it easier to update links globally to keep the Apps consistent if shared though.

packages/theme/src/lib/theme.ts
dmkirshon: No problem, will create the custom `Link` component. My thought was that we would have a set of Typography variants that would align with design to keep consistent with font copy styles, like we do right now with the color palette, so yea basically a global reference stylesheet in a way....

scandycuz (approved):
LGTM

packages/theme/src/lib/theme.ts
scandycuz: I find that modularizing styles using components where possible keeps things a bit more organized.

PR #528, opened 2024-02-14
Implement Portfolio and Transactions tab structure

dmkirshon:
## Update:
- Wallet page to include a Portfolio tab and a Transactions tab for the wallet enhancements and earning claiming

https://projectnewm.atlassian.net/browse/STUD-93

### DIFF
--- a/apps/studio/src/pages/home/wallet/Wallet.tsx
+++ b/apps/studio/src/pages/home/wallet/Wallet.tsx
@@ -63,7 +63,7 @@ const Wallet: FunctionComponent = () => {
             alignItems: "center",
             display: "flex",
             justifyContent: "space-between",
-            marginBottom: 5,
+            mb: 5,
           } }
         >
           <Typography fontWeight={ 800 } variant="h3">
--- a/apps/studio/src/pages/home/wallet/Wallet.tsx
+++ b/apps/studio/src/pages/home/wallet/Wallet.tsx
@@ -1,26 +1,69 @@
-import { FunctionComponent } from "react";
-import { Box, Container } from "@mui/material";
+import { FunctionComponent, ReactNode, SyntheticEvent, useState } from "react";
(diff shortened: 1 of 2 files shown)


apps/studio/src/pages/home/wallet/Wallet.tsx
```diff
@@ -1,26 +1,69 @@
...
     <Container maxWidth={ false }>
-      <Box ml={ [null, null, 3] }>
+      <Box mx={ [null, null, 1.5] }>
         <Box
           sx={ {
             alignItems: "center",
             display: "flex",
             justifyContent: "space-between",
-            mb: 5,
+            marginBottom: 5,
```
escobarjonatan: y u no like mb?

apps/studio/src/pages/home/wallet/Wallet.tsx
```diff
@@ -41,10 +83,49 @@ const Wallet: FunctionComponent = () => {
           ) }
         </Box>
 
-        <UnclaimedRoyalties unclaimedRoyalties={ 0 } />
+        { !!unclaimedRoyalties && (
```
escobarjonatan, line 86: do we want to show something when the opposite is true?

escobarjonatan (approved):

apps/studio/src/pages/home/wallet/Wallet.tsx
dmkirshon: I was playing around with the other padding and margins, just didn't catch it to revert to mb, will fix to not inflict more changes

apps/studio/src/pages/home/wallet/Wallet.tsx
dmkirshon: From what I can tell from the Figma designs, not right now. Will have to look into the other tickets, as this one was just for the tab structure. I will throw the question to design and product but I went for basic appearance for this ticket.

PR #531, opened 2024-02-15
feat: MRKT-11 Create release song component in a shared space. Add Types package for shared interfaces/types

escobarjonatan:
Add:
- Shared song component to be used in marketplace and studio
- New types package to be used for shareable interfaces and types

Notes:
- The purpose of this task was to create the initial song component to be used in studio and marketplace. However because of its dependencies i had to export some logic into a shared space and realized i was getting myself into a rabbit hole the more i moved. I found a good stopping point in order not to make this a huge PR, we will add the playback functionality in a new PR.
- This does not handle the responsiveness of the just released songs in marketing. 

image


apps/marketplace/src/components/Songs.tsx
```diff
@@ -0,0 +1,123 @@
+import { FunctionComponent } from "react";
+import { Stack, Typography } from "@mui/material";
+import { SongCard } from "@newm-web/components";
+/**
+ * TODO: Implement useGetSongsQuery and playback functionality,
+ * see studio/src/pages/home/owners/Songs.tsx
```
escobarjonatan, line 6: This is where i drew the line, the playback and fetch functionality will be completed in a separate PR to keep this one smallish

apps/marketplace/src/components/Songs.tsx
```diff
@@ -0,0 +1,123 @@
...
+          No songs to display at this time.
+        </Typography>
+      ) }
+    </Stack>
+  );
+};
+
+export default Songs;
+
+const tempSongData = [
```
escobarjonatan, line 50: This should be short lived after the next task

nx.json
```diff
@@ -36,6 +36,11 @@
         "{workspaceRoot}/.eslintignore",
         "{workspaceRoot}/eslint.config.js"
       ]
+    },
+    "@nx/js:tsc": {
+      "cache": true,
+      "dependsOn": ["^build"],
+      "inputs": ["production", "^production"]
```
escobarjonatan, line 43: this was added when generating the types package

apps/studio/src/modules/song/types.ts
```diff
@@ -156,69 +152,6 @@ export interface CloudinarySignatureResponse {
   readonly timestamp: number;
 }
 
-export enum MintingStatus {
```
dmkirshon, line 159: Did minting status come over due to the `Song` interface dependence, it won't be something we try and show on Marketplace, correct?

dmkirshon (approved):
LGTM, just a clarifying question on the minting status interface

packages/types/project.json
```diff
@@ -0,0 +1,22 @@
...
+        "tsConfig": "packages/types/tsconfig.lib.json",
+        "assets": ["packages/types/*.md"]
+      }
+    },
+    "lint": {
+      "executor": "@nx/eslint:lint"
+    }
+  },
+  "tags": []
+}
```
scandycuz, line 22: Since it's just a TypeScript library, it may have been possible to not check the box to make it a buildable library (since things will just be imported into another app or package and built there). Making it a buildable libary doesn't impact anything though, it just adds some additional boilerplate code.

scandycuz (approved):
LGTM

apps/studio/src/modules/song/types.ts
escobarjonatan: That is correct, we will not display it, at least from what the mocks say but the type will be reused for song and other things if needed

packages/types/project.json
escobarjonatan: I had the same understanding as you mentioned. However if I make it not buildable you cannot import it to buildable packages/apps. It violates an NX rule. I read online that it is ok to make it buildable even if it's never going to "build"

scandycuz: If it's not possible to import them into other buildable/packages or apps, then that makes sense.

PR #534, opened 2024-02-20
feat: STUD-94 Add empty portfolio and transaction ui state

escobarjonatan:
Portfolio:
image
Transactions:
image

### DIFF
--- a/apps/studio/src/pages/home/wallet/Portfolio.tsx
+++ b/apps/studio/src/pages/home/wallet/Portfolio.tsx
@@ -5,6 +5,7 @@ import { TableSkeleton } from "@newm-web/elements";
 import theme from "@newm-web/theme";
 import { SortOrder } from "@newm-web/types";
 import SongRoyaltiesList from "./SongRoyaltiesList";
+import { EmptyPortfolio } from "./EmptyPortfolio";
 import { useGetUserWalletSongsThunk } from "../../../modules/song";
 
 const Portfolio: FunctionComponent = () => {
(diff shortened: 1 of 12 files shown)


apps/studio/src/pages/home/wallet/Portfolio.tsx
```diff
@@ -52,27 +53,31 @@ const Portfolio: FunctionComponent = () => {
...
+        cols={ windowWidth && windowWidth > theme.breakpoints.values.sm ? 3 : 2 }
+        maxWidth={ maxListWidth }
+        rows={ skeletonRows }
+      />
+    );
+  }
+
+  if (!isLoading && !songs.length) {
+    return <EmptyPortfolio />;
+  }
```
dmkirshon: I noticed, even though I have songs in my portfolio, navigating between transaction tab and portfolio tab, the `EmptyPortfolio` component displays first for a second and then the skeleton appears and then loads my portfolio. Should probably go straight to skeleton and then to portfolio.

apps/studio/src/pages/home/wallet/Portfolio.tsx
dmkirshon, line 1: For our future note, we can let Denny know to add a Portfolio task or add a bug ticket, I noticed that a lot of data is being used to load the song album covers on the portfolio even though they are tiny. For reference 8 songs ended up being 24.5MB of data. 

image

dmkirshon (changes requested):
One minor change for order of Components loading.

apps/studio/src/pages/home/wallet/Portfolio.tsx
escobarjonatan: ahh i see, fixed it by adding isSuccess to the custom thunk we have for this. Added isError in case we need it since i was there already

apps/studio/src/pages/home/wallet/Portfolio.tsx
escobarjonatan: noted, this is interesting bc we should be getting those images in the correct size/format so those other images that are loading may be coming from somewhere else or something is going on.

dmkirshon (approved):
LGTM

PR #535, opened 2024-02-20
feat: STUD-96 Add UI for no pending earnings in wallet

escobarjonatan:
image

### DIFF
--- a/apps/studio/src/pages/home/wallet/NoPendingEarnings.tsx
+++ b/apps/studio/src/pages/home/wallet/NoPendingEarnings.tsx
@@ -19,7 +19,7 @@ export const NoPendingEarnings: FunctionComponent = () => (
       <Typography color={ theme.colors.green } fontWeight={ 500 }>
         No pending earnings to claim
       </Typography>
-      <Typography color="grey200" fontWeight={ 500 }>
+      <Typography color={ theme.colors.grey200 } fontWeight={ 500 }>
         Total earnings accrued so far: ##.##Ɲ (~$#.##)
       </Typography>
     </Stack>
--- a/apps/studio/src/pages/home/wallet/NoPendingEarnings.tsx
+++ b/apps/studio/src/pages/home/wallet/NoPendingEarnings.tsx
@@ -1,0 +1,27 @@
+import { FunctionComponent } from "react";
+import { Stack, Typography } from "@mui/material";
+import { CheckCircle } from "@newm-web/assets";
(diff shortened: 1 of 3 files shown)


apps/studio/src/pages/home/wallet/Wallet.tsx
```diff
@@ -82,8 +91,10 @@ const Wallet: FunctionComponent = () => {
           ) }
         </Box>
 
-        { !!unclaimedRoyalties && (
+        { unclaimedRoyalties ? (
           <UnclaimedRoyalties unclaimedRoyalties={ unclaimedRoyalties } />
+        ) : (
+          <NoPendingEarnings />
         ) }
```
dmkirshon: The actual conditional, per the Figma, should probably be if no songs have been uploaded (minted? might need to verify with Product?) -> then don't show either box  -> otherwise if song has been uploaded -> then unclaimedRoyalties conditional.

 Can be added to either of your PRs or I can modify on my STUD-95 PR.

apps/studio/src/pages/home/wallet/NoPendingEarnings.tsx
```diff
@@ -0,0 +1,27 @@
...
+      p: 2,
+      width: "fit-content",
+    } }
+  >
+    <CheckCircle fill={ theme.colors.green } />
+    <Stack>
+      <Typography color={ theme.colors.green } fontWeight={ 500 }>
+        No pending earnings to claim
+      </Typography>
+      <Typography color="grey200" fontWeight={ 500 }>
```
dmkirshon: Text color is white due to theme reference missing, per figma should be grey200 in UI.

```suggestion
      <Typography color={ theme.colors.grey200 } fontWeight={ 500 }>
```

dmkirshon (changes requested):
Minor font color update.

apps/studio/src/pages/home/wallet/NoPendingEarnings.tsx
escobarjonatan: 🤦🏽‍♂️

apps/studio/src/pages/home/wallet/Wallet.tsx
escobarjonatan: Confirmed with denny and you are correct, we should only show either or based on if the have songs or not. updated the logic to reflect that

dmkirshon (approved):
LGTM

PR #547, opened 2024-03-02
Add search ui to marketplace

escobarjonatan:
# Add
- Search UI for marketplace

# Update
-  text input width prop `widthType` it was getting ignored by a change inside the input component
- styled-components dependency. This was needed in order for styles to load properly in our marketplace app. See video for issue.

Demo:
https://github.com/projectNEWM/newm-web/assets/11530120/05589f04-9892-4cd9-b888-a199d19b6b86

Styles issue on initial load:
https://github.com/projectNEWM/newm-web/assets/11530120/e293c574-146e-4a6a-b7d3-e5afd3e0d886


apps/marketplace/src/common/hoc.tsx
```diff
@@ -0,0 +1,35 @@
+"use client";
```
scandycuz: Would be good to give this component a more specific name, e.g. `StyledComponentsRegistry`. It could also go in the `components` directory.

scandycuz (commented):
LGTM, but had one comment. Also, it looks like the CI process failed.

apps/marketplace/src/common/hoc.tsx
escobarjonatan: i thought about it, i even thought it could go in the `app` directory since it's a 1 time use component utility that will only be used in the layout and no where else. Adding it to the components i thought would give it more of that "reusability" feel when it's really not. I dont mind moving it though

escobarjonatan:
> LGTM, but had one comment. Also, it looks like the CI process failed.

yea, i thought it was happening bc of the node environment inconsistency between my machine and the pipeline. I created this PR to ensure we are all on the same version to prevent issues: https://github.com/projectNEWM/newm-web/pull/549

apps/marketplace/src/common/hoc.tsx
scandycuz: The `app` directory is actually a special directory used for the Next.js [App router](https://nextjs.org/docs/app/building-your-application/routing), so it's probably best to not have it there.

In retrospect, the `common` directory may actually be a good place for it. It looks like there are components there. Renaming it to be more specific is definitely good though.

apps/marketplace/src/components/header/Search.tsx
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  const initialValues = {
+    searchTerm,
+  };
+
+  const handleSearch = ({ searchTerm }: FormikValues) => {
+    if (!searchTerm) return;
+
+    router.push(`/search?searchTerm=${searchTerm}`);
```
scandycuz, line 25: Probably not worth changing, but I think `query` is a pretty common param name for a search parameter.

scandycuz (approved):

PR #549, opened 2024-03-03
chore: Updating pipelines and scripts to use the same node version

escobarjonatan:
Add consistent node version in app and pipelines to reduce potential versioning issues

After this gets merged, i recommend to delete `node_modules` folder and install your dependencies. Ideally if you are not modifying the package.json in your pr you should do `npm ci`  in order to install the dependencies from the package-lock . Also, if you are using a node version older than `18.19.0` it will throw an error and you will need to upgrade to that. Using nvm or managing versions you can do: `nvm use 18.19.0` if version not installed it will give you the command to install it


.github/workflows/firebase-hosting-deploy-published-release.yml
```diff
@@ -8,7 +8,13 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - run: npm install
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: .nvmrc
+          cache: npm
+
+      - run: npm ci
```
escobarjonatan, line 17: instead of `npm i`, adding `npm ci`, it ensures consistent and fast dependency installations from package-lock.json without modifying it, ideal for CI/CD pipelines, while npm install updates dependencies and the lock file

.nvmrc
```diff
@@ -0,0 +1 @@
+18.19.0
```
escobarjonatan, line 1: version can be controlled here that can be used in the pipeline and preinstalled script

scandycuz (approved):

PR #550, opened 2024-03-03
MRKT-26: artist spotlight 

scandycuz:
Adds "Artist spotlight" section (with static content) to home page.

### Also

Swaps `getResizedAlbumCoverImageUrl` for `resizeCloudinaryImage`, which works for all Cloudinary image urls.


apps/marketplace/src/components/index.ts
scandycuz, line 1: Removed this file, as we weren't using it, and I think keeping it updated adds more work than it is useful for the components directory.

apps/marketplace/src/components/Artist.tsx
```diff
@@ -0,0 +1,69 @@
...
+          variant="h4"
+        >
+          { subtitle }
+        </Typography>
+      </Stack>
+    </Stack>
+  );
+};
+
+export default Artist;
```
scandycuz, line 69: This component can also be used for the "Similar artists" section, by passing the `orientation="row"` prop.

apps/marketplace/src/components/Artist.tsx
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Image and information for an artist. Can be vertically
+ * or horizontally aligned.
+ */
+const Artist: FunctionComponent<ArtistProps> = ({
+  imageUrl,
+  title,
+  isLoading,
+  subtitle,
+  orientation,
```
escobarjonatan, line 25: maybe this can be optional and we can set the default to "column" or whatever it is used most from the mocks

apps/marketplace/src/components/ArtistSpotlight.tsx
```diff
@@ -0,0 +1,104 @@
...
+    }, 1000);
+  }, []);
+
+  return (
+    <Stack mb={ 4 }>
+      <Stack alignItems="center" mb={ 5 } mt={ 20 }>
+        <Typography variant="h3">ARTIST SPOTLIGHT</Typography>
+      </Stack>
+
+      <Grid justifyContent="center" container>
```
escobarjonatan: What do you think about removing/reducing the spacing in the artist card itself and instead apply it through this container through column/row gaps. That way the user's clickable area has spacing/transition between artists, for example adding `columnGap={2}` here and removing the `pr={2}` in line 34 can add some clickable breathing room:


image

apps/marketplace/src/components/ArtistSpotlight.tsx
```diff
@@ -0,0 +1,104 @@
...
+  useEffect(() => {
+    setTimeout(() => {
+      setIsLoading(false);
+    }, 1000);
+  }, []);
+
+  return (
+    <Stack mb={ 4 }>
+      <Stack alignItems="center" mb={ 5 } mt={ 20 }>
+        <Typography variant="h3">ARTIST SPOTLIGHT</Typography>
```
escobarjonatan:
```suggestion
           <Typography fontSize={ ["24px", "24px", "32px"] } variant="h3">
            ARTIST SPOTLIGHT
          </Typography>
```
to align with the just released tile

apps/marketplace/src/components/index.ts
escobarjonatan, line 1: I dont agree that we should delete this, it's useful when importing/exporting things instead of having longer import paths. I think we should keep it

apps/marketplace/src/components/ArtistSpotlight.tsx
escobarjonatan: also, noticed that these are "clickable" but not tabbable. Assuming the action here is to navigate to the artist page maybe the "clickable" look/feel is the artist name and we can forward the click from the card to the artist name. This way it feels like the card is clickable but the artist name is what's actionable. Last resort, you just do tabIndex=0 and role="button" here that will be a better ux for keyboard users

apps/marketplace/src/components/Artist.tsx
scandycuz: It's currently only used in two sections, the "Artist spotlight", and the "Similar artists", each with a different layout, so I wasn't sure it made sense to give it a default.

apps/marketplace/src/components/index.ts
scandycuz: That works. We should update the code to import from the top level of `components` then though.

apps/marketplace/src/components/ArtistSpotlight.tsx
scandycuz: I was initially using the `spacing` prop, in the [docs](https://mui.com/material-ui/api/grid/), but that was adding some negative margins above and below the content. The `gap` prop seems to work though.

scandycuz: I think it's a better user experience to make more than just the title clickable. I was going to go with just wrapping the "Artist" item in an unstyled button. It could be good to talk to UX about some sort of hover styling as well.

apps/marketplace/src/components/index.ts
scandycuz: Re-added the file and updated the imports.

apps/marketplace/src/components/ArtistSpotlight.tsx
escobarjonatan: Nice updates.

I wasn't trying to say to not make the whole thing clickable, but more something along the lines of forwarding the click of the outer div(card) to the title which would be a link but i like what you did

escobarjonatan (approved):

PR #554, opened 2024-03-07
STUD-184: collaborator list display bug

scandycuz:
Fixes a bug where a collaborator would appear in the collaborator list as not having created an account if they hadn't added a profile image.

### Also

- Optimized various image sizes


apps/studio/src/pages/home/library/EditSong.tsx
```diff
@@ -282,7 +283,7 @@ const EditSong: FunctionComponent = () => {
         <ProfileImage
           alt="Song cover art"
           height="90px"
-          src={ coverArtUrl }
+          src={ resizeCloudinaryImage(coverArtUrl, { height: 180, width: 180 }) }
```
scandycuz, line 286: The images were a bit blurry if they were resized to the actual dimensions. I made them a bit larger so they look clear.

escobarjonatan (approved):

PR #559, opened 2024-03-13
STUD-145: Update Portfolio table and time period filters for viewing accumulated royalties

dmkirshon:
Added:
- Test utils to mock Royalty Earnings and their creation dates
- Filtering functionality based on Earnings creation dates
- Sorting of Earnings based on summation of all Earnings for a song
- Saved Period Filter option between `Portfolio` and `Transactions` tabs, resets on page navigation

Updated:
- Filter drop down arrow to be always visible
- Filter hover/selected option color
- Filter option copy
- Default no royalty earnings from `--.--` to Ɲ currency formatting
- Pagination to reflect Earnings order instead of pagination received from Wallet

Refactored:
- Royalty Earnings Table row color code
- `TableDropdownSelect.tsx` to remove references specific to `SongRoyaltiesList.tsx`

Notes:
- Test Royalty Earnings data is based on length of song title received from wallet 
  - Odd song title lengths have 0 Earnings

https://projectnewm.atlassian.net/browse/STUD-145


escobarjonatan (approved):

apps/studio/src/pages/home/wallet/Portfolio.tsx
```diff
@@ -30,16 +35,21 @@ const Portfolio: FunctionComponent = () => {
     { data: walletSongsResponse, isLoading, isSuccess },
   ] = useGetUserWalletSongsThunk();
 
-  const songs =
-    walletSongsResponse?.data?.songs?.map((entry) => entry.song) || [];
+  const songs = useMemo(
+    () => walletSongsResponse?.data?.songs?.map((entry) => entry.song) || [],
+    [walletSongsResponse?.data?.songs]
+  );
```
scandycuz, line 41: Another way to do this would be to use the [transformResponse](https://redux-toolkit.js.org/rtk-query/usage/customizing-queries#customizing-query-responses-with-transformresponse) function that RTKQuery provides. That way the logic to fetch the data and transform the response would all be in the same place.

scandycuz (commented):
Had one comment, but not necessarily worth delaying the PR for.

apps/studio/src/pages/home/wallet/Portfolio.tsx
```diff
@@ -55,7 +65,40 @@ const Portfolio: FunctionComponent = () => {
...
+          totalRoyaltyAmount,
+        };
+      })
+      // Sort filtered songs by descending totalRoyaltyAmount
+      .sort((a, b) => b?.totalRoyaltyAmount - a?.totalRoyaltyAmount)
+      // handle pagination for song earnings
+      .slice(pageIdx * skeletonRows, pageIdx * skeletonRows + rowsToRender);
+
+    setWalletSongsRoyaltyCombined(combinedEarnings);
+  }, [pageIdx, skeletonRows, songs, walletPortfolioTableFilter, windowHeight]);
```
scandycuz, line 97: Ideally, we could move the sorting, pagination, and potentially royalty calculations to the back-end.

dmkirshon: Yes, that would be the ideal with the endpoint, it would probably require us to feed the backend with an array of songs that are provided from the wallet. Otherwise this logic will still need to be consolidated on the frontend to deal with a single song earnings query, for summation of earnings, creation date filtering, and pagination. I will bring it up to the backend team as they start getting into building out the query.

apps/studio/src/pages/home/wallet/Portfolio.tsx
dmkirshon: I will leave this till we have to do the endpoint query update for the Earnings table and we can just make note of it in the task that gets created for it.

PR #561, opened 2024-03-14
STUD-180 Implement cover art agreement checkbox

escobarjonatan:
image

### DIFF
--- a/apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
+++ b/apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
@@ -1,7 +1,8 @@
-import { Box, Stack, useTheme } from "@mui/material";
+import { Box, Link, Stack, useTheme } from "@mui/material";
 import {
   Alert,
   Button,
+  CheckboxField,
   DropdownMultiSelectField,
   DropdownSelectField,
   ErrorMessage,
@@ -24,7 +25,7 @@ import { useConnectWallet } from "@newm.io/cardano-dapp-wallet-connector";
 import { useFormikContext } from "formik";
 import { FunctionComponent, useEffect, useRef, useState } from "react";
 import { useNavigate, useParams } from "react-router-dom";
-import { useAppDispatch } from "../../../common";
+import { NEWM_STUDIO_FAQ_URL, useAppDispatch } from "../../../common";
 import { PlaySong, PricingPlansDialog } from "../../../components";
 import SelectCoCeators from "../../../components/minting/SelectCoCreators";
 import {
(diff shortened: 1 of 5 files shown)


apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
```diff
@@ -108,9 +116,13 @@ const BasicSongDetails: FunctionComponent<BasicDonDetailsProps> = ({
     }
   }, [isArtistPricePlanSelected, isPricingPlansOpen, setFieldValue]);
 
+  const hasCoverArtChanged = values.coverArtUrl !== initialValues.coverArtUrl;
```
dmkirshon, line 123: Is this logic, and the `useEffect` on line `164` below, based on the implication that if they have selected the checkbox and then replace their cover art, that the checkbox will reset each time they change their cover art? 

Right now it does the following:
1. Add image to field
2. Select checkbox to agree
3. Upload new/different image to field
4. Checkbox is still checked

dmkirshon (changes requested):
LGTM, just one question about logic flow on cover art field updates

apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
escobarjonatan: imo this is ok, we reset the status if it's not the same as it initially was. if they change it afterwards it doesn't matter since they've read the condition initially and agreed to it? i'll ask denny to check it out and see what he thinks of the flow

scandycuz: I think either way could work. Resetting the box for each image would be more accurate, but the user may not notice that the box has been unchecked when they upload a new image and have to go through the process of seeing the error and resubmitting. If we leave it checked, that would be more convenient, and as Jonatan stated, could work because the user has confirmed they're aware of the guidelines.

dmkirshon: Sounds good, as part of that initial question, I just wanted to understand if the intention was specifically for every time the field was touched or just when the song was already uploaded but still in edit mode within the Library. 

If once they checkmark the box, it's seen as acknowledged, I'm good with keeping the logic as is. With the way the logic flows they still have to acknowledge the guidelines again if they update the cover art in an upload/re-submit mode anyway.

dmkirshon (approved):
I don't believe the logic change for re-acknowledging the cover art message when the user switches the cover art multiple times on initial Song upload should be a blocker for pushing this PR through, if required by Product another task to update the logic can be created to address this UX.

PR #566, opened 2024-03-26
STUD-157: Add instrumental song option and refactor types to avoid confusion

escobarjonatan:
View details:
image

Advanced details:
image

### DIFF
--- a/apps/studio/src/modules/song/types.ts
+++ b/apps/studio/src/modules/song/types.ts
@@ -68,6 +68,7 @@ export interface Collaborator {
  */
 export interface PostSongRequest {
   readonly album?: string;
+  readonly archived?: boolean;
   readonly barcodeNumber?: string;
   readonly barcodeType?: string;
   readonly compositionCopyrightOwner?: string;
@@ -76,6 +77,7 @@ export interface PostSongRequest {
   readonly coverRemixSample?: boolean;
   readonly description?: string;
   readonly genres: ReadonlyArray<string>;
+  readonly instrumental?: boolean;
   readonly ipis?: ReadonlyArray<string>;
   readonly isrc?: string;
   readonly iswc?: string;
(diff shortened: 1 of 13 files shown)


apps/studio/src/modules/song/types.ts
```diff
@@ -91,9 +93,22 @@ export interface PostSongRequest {
...
-  readonly agreesToCoverArtGuidelines?: boolean;
+/**
+ * PatchSong as stored in the NEWM API.
+ */
+export interface PatchSongRequest
+  extends Omit<PostSongRequest, "title" | "genres"> {
+  readonly genres?: ReadonlyArray<string>;
+  readonly id: string;
+  readonly title?: string;
+}
```
escobarjonatan, line 104: The idea behind refactoring types a bit was to maintain the API isolated from the thunks and the properties we needed for UI's internal logic.

apps/studio/src/modules/song/types.ts
```diff
@@ -91,9 +93,22 @@ export interface PostSongRequest {
...
+
+/**
+ * Type for thunk request to upload a song.
+ * It does not have to be the same as the API request.
+ */
+export interface UploadSongThunkRequest
```
escobarjonatan, line 110: By keeping the thunk and endpoint interface separated it allows us to have different properties, like in this case, the endpoint has "instrumental" but for our thunk we use "isInstrumental" to remain consistent with our boolean naming strategy.

apps/studio/src/modules/song/types.ts
```diff
@@ -103,13 +118,38 @@ export interface UploadSongRequest
...
+      | "ipi"
+      | "isExplicit"
+      | "isInstrumental"
+      | "isMinting"
+      | "owners"
+      | "stageName"
+    > {
+  readonly mintingStatus?: MintingStatus;
+  readonly shouldRedirect?: boolean;
+}
```
escobarjonatan, line 151: Similar to the post, i separated the patch a bit but reused some properties we can always add/remove as needed

scandycuz (approved):

apps/studio/src/modules/song/types.ts
```diff
@@ -103,13 +118,38 @@ export interface UploadSongRequest
...
+ * It does not have to be the same as the API request.
+ */
+export interface PatchSongThunkRequest
+  extends Omit<PatchSongRequest, "instrumental" | "coverArtUrl">,
+    Pick<
+      UploadSongThunkRequest,
+      | "artistName"
+      | "companyName"
+      | "consentsToContract"
+      | "coverArtUrl"
```
dmkirshon, line 139: ~~Am I missing something on why `coverArtUrl` was omitted in the `PatchSongRequest` but picked from the `UploadSongThunkRequest`?~~

Understand now why `coverArtUrl` is addressed like this, it can be a type of `string` or `File`. That is a bit of a head turner in use.

dmkirshon (approved):
LGTM, just got lost for a second with the types.

And to verify there is no business logic currently needed for them to select the genre as instrumental even if they select the toggle, correct? Or even that it autotoggles if they select the genre on the Basic Details page?

PR #567, opened 2024-03-26
STUD-166: Add tooltip to song error states in library

escobarjonatan:
Declined:
image
Error:
image

### DIFF
--- a/apps/studio/src/pages/home/library/MintingStatus.tsx
+++ b/apps/studio/src/pages/home/library/MintingStatus.tsx
@@ -34,6 +34,17 @@ const UI_MINTING_STATUS: Record<MintingStatusType, string> = {
   Undistributed: "Undistributed",
 };
 
+// Indicates all "An error occurred" statuses. Used externally to determine tooltip content.
+export const ErrorOccurredMintingStatuses = [
+  MintingStatusType.ArweaveUploadException,
+  MintingStatusType.DistributionException,
+  MintingStatusType.MintingException,
+  MintingStatusType.MintingPaymentException,
+  MintingStatusType.MintingPaymentTimeout,
+  MintingStatusType.ReleaseCheckException,
+  MintingStatusType.SubmittedForDistributionException,
+];
+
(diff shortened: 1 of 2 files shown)


apps/studio/src/pages/home/library/MintingStatus.tsx
```diff
@@ -34,6 +34,17 @@ const UI_MINTING_STATUS: Record<MintingStatusType, string> = {
...
+// Indicates all "An error occurred" statuses. Used externally to determine tooltip content.
+export const ErrorOccurredMintingStatuses = [
+  MintingStatusType.ArweaveUploadException,
+  MintingStatusType.DistributionException,
+  MintingStatusType.MintingException,
+  MintingStatusType.MintingPaymentException,
+  MintingStatusType.MintingPaymentTimeout,
+  MintingStatusType.ReleaseCheckException,
+  MintingStatusType.SubmittedForDistributionException,
+];
```
escobarjonatan, line 46: Added this here in order to keep minting ui status in one place so when we update the `UI_MINTING_STATUS` we can update this if needed.

scandycuz (approved):

apps/studio/src/pages/home/library/SongList.tsx
```diff
@@ -31,7 +32,8 @@ import {
   SortOrder,
 } from "@newm-web/types";
 import { useNavigate } from "react-router-dom";
-import { MintingStatus } from "./MintingStatus";
+import { min } from "moment";
```
dmkirshon: This should probably be removed, don't see the reference being used.

dmkirshon (changes requested):
LGTM, just need to remove superfluous import.

apps/studio/src/pages/home/library/SongList.tsx
escobarjonatan: gotta love auto import as you type 🫠

dmkirshon (approved):
LGTM

PR #568, opened 2024-03-26
STUD-190: Update support nav copy

escobarjonatan:


apps/studio/src/pages/home/owners/OwnersTable.tsx
```diff
@@ -212,7 +212,7 @@ export default function OwnersTable({
                       `${firstName} ${lastName}`
                     ) : (
                       <Typography fontStyle="italic" fontWeight={ 400 }>
-                        Waiting on account creation
+                        Waiting on collaborator&apos;s acceptance
```
scandycuz, line 215: 👍

scandycuz (approved):

apps/studio/src/pages/home/owners/OwnersTable.tsx
```diff
@@ -212,7 +212,7 @@ export default function OwnersTable({
                       `${firstName} ${lastName}`
                     ) : (
                       <Typography fontStyle="italic" fontWeight={ 400 }>
```
dmkirshon, line 214: This italics is a bit strange and seems out of place to other font styles we've used... can't find any reference to it on Figma either.

dmkirshon (approved):
LGTM, just comment about UI font style

apps/studio/src/pages/home/owners/OwnersTable.tsx
escobarjonatan: Looks like the mocks are also different and dont include either scenario. We can do that as a separate update when we add icons as part of collaborator status

PR #570, opened 2024-03-26
MRTK-22: artist banner

scandycuz:
- Adds a banner to the Marketplace artist page
- Refactors the back header and profile banner as shared components


apps/marketplace/src/app/page.tsx
```diff
@@ -1,14 +1,18 @@
...
+import { ArtistSpotlight, Header, Songs } from "../components";
 
 const Home: FunctionComponent = () => {
   return (
-    <Container sx={ { flexGrow: 1 } }>
-      <Songs />
-      <ArtistSpotlight />
-    </Container>
+    <Stack direction="column">
+      <Header />
```
escobarjonatan: So we don't want this in all of the pages? 🤔

escobarjonatan (approved):

apps/marketplace/src/app/page.tsx
scandycuz: I was wondering the same thing, but it's not in the mock-ups. I'll reach out to product to confirm.

PR #573, opened 2024-03-27
STUD-145: Add Portfolio end of table copy

dmkirshon:
Add:
- "Caught up" copy at the end of the Wallet Portfolio table

image
image

### DIFF
--- a/apps/studio/src/pages/home/wallet/SongRoyaltiesList.tsx
+++ b/apps/studio/src/pages/home/wallet/SongRoyaltiesList.tsx
@@ -16,6 +16,7 @@ import { Song } from "@newm-web/types";
 import { Dispatch, SetStateAction } from "react";
 import { TablePagination, Typography } from "@newm-web/elements";
 import currency from "currency.js";
+import AllCaughtUp from "./AllCaughtUp";
 import { selectUi, setWalletPortfolioTableFilter } from "../../../modules/ui";
 import {
   TableDropdownMenuParameters,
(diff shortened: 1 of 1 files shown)


apps/studio/src/pages/home/wallet/SongRoyaltiesList.tsx
```diff
@@ -74,6 +75,7 @@ export default function SongRoyaltiesList({
   const dispatch = useAppDispatch();
 
   const TABLE_WIDTH = 700;
+  const isLastRowOnPageVisible = lastRowOnPage >= totalCountOfSongs;
```
escobarjonatan, line 78: Will the > ever be true? 👀

escobarjonatan (approved):

apps/studio/src/pages/home/wallet/SongRoyaltiesList.tsx
dmkirshon: I didn't look deeper into the pagination logic but the `lastRowOnPage` seems to calculate `page` * `rowsPerPage`, so it is either greater than `totalCountOfSongs` or equal. 

`const lastRowOnPage = pageIdx * rowsPerPage + rowsPerPage;`

Should I fix that last part to be `+ rows`? And correct the `isLastRowOnPageVisible` logic to `===`?

dmkirshon: I just double checked and the `page` * `rowsPerPage` logic is the same on the `Library` page as well.

PR #575, opened 2024-03-28
STUD-203: Reset consentsToContract field to false when navigating back

escobarjonatan:


apps/studio/src/pages/home/uploadSong/ConfirmAgreement.tsx
```diff
@@ -32,6 +32,13 @@ const ConfirmAgreement: FunctionComponent<ConfirmAgreementProps> = ({
     }
   };
 
+  // Reset consent to contract when unmounting
+  useEffect(() => {
+    return () => {
+      setFieldValue("consentsToContract", false);
+    };
+  }, [setFieldValue]);
```
escobarjonatan, line 40: This is needed in order to set `consentsToContract` to false after approving it -> navigate back to advanced details -> navigate forward to the confirm page. It was saving the previous `true` value

scandycuz (approved):

PR #579, opened 2024-03-30
[MRKT-24, MRKT-25]: Artist detail page

scandycuz:
- Completes the artist detail page UI with the artist's songs and similar artists

### Also 

- Refactors song card to make it easier to interact with and improve responsiveness
- Various minor refactors


apps/marketplace/src/app/item/[songId]/page.tsx
```diff
@@ -22,26 +21,22 @@ const SingleSong: FunctionComponent<SingleSongProps> = ({ params }) => {
     }, 1000);
   }, []);
 
-  if (isLoading) {
-    return <ItemSkeleton />;
-  }
-
```
Comment (dmkirshon) [inline apps/marketplace/src/app/item/[songId]/page.tsx]: I understand you refactored the SongCard to expose the isLoading prop, but just wanted to explain why I originally added the `ItemSkeleton`. 

When all the remaining components/pieces are added, the Item page will have to have data fetched for the description, pricing and other metadata; which will probably be separate queries. I saw it as let all those pieces load together before rendering, instead of individual skeletons for each component. 

Do you see it as a piecewise load, or some can load first while other data is still being queried?

packages/components/src/lib/SongCard.tsx
```diff
@@ -1,225 +1,221 @@
...
-  }, [theme.breakpoints.values.md, windowWidth]);
+  const handleCardClick = (event: MouseEvent | KeyboardEvent) => {
+    event.preventDefault();
+    event.stopPropagation();
+    onCardClick?.();
+  };
 
-  const handlePriceClick = () => {
-    onPriceClick?.();
+  const handlePlayPauseClick = (event: MouseEvent | KeyboardEvent) => {
```
dmkirshon, line 49: When I click the play button from the `Just Released` listings it navigates me to the Item page. Not sure why on initial review, I think it has to do with the `Clickable` component being the parent or might be both click actions competing and the top level winning so the `stopPropagation` isn't actually stopping the child/parent relationship.

packages/components/src/lib/SongCard.tsx
```diff
@@ -1,225 +1,221 @@
...
+        <Stack alignItems="center" justifyItems="center" position="relative">
+          <Box
+            aria-label="Song cover art"
+            component="img"
+            height="100%"
+            src={
+              coverArtUrl
+                ? resizeCloudinaryImage(coverArtUrl, {
+                    height: 400,
+                    width: 400,
```
dmkirshon: Should this be adjusted based on the sizing or 400px covers the basis for the small and large sizes?

dmkirshon (changes requested):
Overall LGTM, but had a few questions and ran into one click propagation issue.

packages/components/src/lib/SongCard.tsx
dmkirshon: After reviewing it again, once I add `onPlayPauseClick={ () => {} }` to the `SongCard`  component in `Songs` , the `stopPropagation` functions correctly. Was just missing the event listener on the play/pause button to trigger instead of only having the card listener.

dmkirshon (approved):
LGTM, still have questions above but resolved what I thought was a blocker on the interactive event listener propagation.

apps/marketplace/src/app/item/[songId]/page.tsx
Comment (scandycuz) [reply apps/marketplace/src/app/item/[songId]/page.tsx]: I see what your saying, it should have its own skeleton component that loads all at once. I'll revert the change.

packages/components/src/lib/SongCard.tsx
scandycuz: Good catch. Technically, there shouldn't be a use case where a song has a stream url but doesn't have an `onPlayPauseClick` prop, but I'll update the component to allow for that use case, just to be complete.

packages/components/src/lib/SongCard.tsx
scandycuz: Yeah, that should probably be updated to be slightly larger than `400px` for the larger version, as the images appear somewhat pixelated when they are the actual size of what is rendered.

packages/elements/src/lib/skeleton/TableSkeleton.tsx
```diff
@@ -57,3 +57,5 @@ export const TableSkeleton: FunctionComponent<TableSkeletonProps> = ({
     </Box>
   );
 };
+
+export default TableSkeleton;
```
escobarjonatan, line 61: Is this change just a preference thing?

scandycuz: Yeah, almost of the components are exported using a default export, so I think it's good to be consistent. I also like it because it helps denote what the primary export of a file is.

PR #586, opened 2024-04-04
STUD-150: Update submit ticket link

escobarjonatan:


dmkirshon (approved):
LGTM, just want to make sure that the link will become live when we do our next release

escobarjonatan:
It’s already available, you cannot access the ticket support form url with a newm account, open it in an incognito window

PR #592, opened 2024-04-15
MOBILE-192: Add new app to connect mobile app to web wallet

escobarjonatan:
https://github.com/projectNEWM/newm-web/assets/11530120/05e06277-feb2-485a-93b0-4bc305a59752


apps/mobile-wallet-connector/project.json
```diff
@@ -0,0 +1,58 @@
...
+      "defaultConfiguration": "development",
+      "options": {
+        "buildTarget": "mobile-wallet-connector:build",
+        "dev": true
+      },
+      "configurations": {
+        "development": {
+          "buildTarget": "mobile-wallet-connector:build:development",
+          "dev": true,
+          "port": 3000
```
escobarjonatan: TODO: This needs to be set to `4300` there are currently CORS issues that need to get addressed

apps/mobile-wallet-connector/src/api/utils.ts
```diff
@@ -0,0 +1,89 @@
+import { BaseQueryApi } from "@reduxjs/toolkit/dist/query/baseQueryTypes";
```
escobarjonatan, line 1: File mostly copied from studio, wanted to keep them separate to keep control

apps/mobile-wallet-connector/src/app/connect-wallet/connected/page.tsx
```diff
@@ -0,0 +1,21 @@
+"use client";
+import { FunctionComponent } from "react";
+import { Typography } from "@mui/material";
+
+// TODO: Page will not be needed as there is no way to determine mobile connectivity
```
escobarjonatan: TODO: Page will not be needed as there is no way to determine mobile connectivity

apps/mobile-wallet-connector/src/components/ConnectWallet.tsx
```diff
@@ -0,0 +1,98 @@
+import currency from "currency.js";
```
escobarjonatan, line 1: File mostly copied from studio, wanted to keep them separate to keep control

apps/studio/src/components/Toast.tsx
```diff
@@ -1,86 +1,24 @@
-import { SyntheticEvent, useEffect, useState } from "react";
-import { Alert, Snackbar, Typography, useTheme } from "@mui/material";
-import { CheckCircle, CloseCircleFill } from "@newm-web/assets";
+import { Alert } from "@newm-web/components";
```
escobarjonatan, line 1: Moved the toast to a shared place so it can be reused in other apps, kept app specific logic here

apps/mobile-wallet-connector/src/api/constants.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Maps RTKQuery API endpoint names with name that
+ * back-end expects for recaptcha action argument.
+ */
+export const recaptchaEndpointActionMap: Record<string, string> = {
+  answerChallenge: "answer_challenge",
+  generateChallenge: "generate_challenge",
+  getQRCode: "qrcode",
+};
```
escobarjonatan, line 9: Thought about creating one reusable "util" that combines this map and current studio map related to recaptcha but didn't want to add unnecessary keys we weren't going to use in this app, we might in marketplace though... we can cross that bridge when we get there or now if you all like 🤷🏽‍♂️

apps/mobile-wallet-connector/src/app/connect-wallet/page.tsx
```diff
@@ -0,0 +1,88 @@
...
+        { ({ isSubmitting }) => (
+          <Form>
+            <Stack alignItems="center" px={ [2, 2, 0] } rowGap={ 3.5 }>
+              <Button
+                disabled={ isFormSubmitDisabled || isSubmitting }
+                isLoading={ isConnectFromMobileLoading }
+                type="submit"
+                width="compact"
+              >
+                Connect from Mobile app
```
dmkirshon, line 70: I haven't seen the whole flow from the mobile app side, but "Connect from Mobile app" seems like strange copy for this button. Maybe like "Generate QR code for Mobile app"?

apps/mobile-wallet-connector/src/app/connect-wallet/scan-code/page.tsx
```diff
@@ -0,0 +1,157 @@
...
+    }
+  }, [isCopied]);
+
+  return (
+    <>
+      <Typography component="h1" fontWeight="700" mt={ 4 } variant="body2">
+        Scan the QR Code with Newm app
+      </Typography>
+      <Typography mb={ 5 } mt={ 0.5 } variant="subtitle1">
+        or paste the code in the app
```
dmkirshon: This and the label copy above the actual text code seem redundant, more of a UX gripe

apps/mobile-wallet-connector/src/app/connect-wallet/scan-code/page.tsx
```diff
@@ -0,0 +1,157 @@
...
+          value={ connectionId }
+        />
+      ) : (
+        <Skeleton height={ 280 } variant="rectangular" width={ 280 } />
+      ) }
+      <Stack mt={ 3 }>
+        { timeLeft ? (
+          <Typography>
+            { timeLeft.total <= 0 || isNaN(timeLeft.total)
+              ? "The code has expired. Please refresh the page to generate a new code."
```
dmkirshon: Does this still make sense if the page is redirected from the QR code back to the connect-wallet page anyway?

apps/mobile-wallet-connector/src/components/ConnectWallet.tsx
dmkirshon: I understand the API query separation of control, not too sure why this one is separate? Isn't the basics of this component needed in all three apps?

apps/mobile-wallet-connector/src/app/connect-wallet/scan-code/page.tsx
```diff
@@ -0,0 +1,157 @@
...
+      setTimeout(() => {
+        setIsCopied(false);
+      }, 2000);
+    }
+  }, [isCopied]);
+
+  return (
+    <>
+      <Typography component="h1" fontWeight="700" mt={ 4 } variant="body2">
+        Scan the QR Code with Newm app
```
dmkirshon: NEWM instead of Newm

dmkirshon (approved):
LGTM, well done, just a few comments spaced around

apps/mobile-wallet-connector/src/components/ConnectWallet.tsx
escobarjonatan: We need to keep these separate for things like dispatching actions or running any other app specific logic we might need. If at a later time we find these will be the same we can work on a shareable solution.

apps/mobile-wallet-connector/src/app/connect-wallet/page.tsx
escobarjonatan: It does more than just generate the QRCode in the background like request and answer a challenge. I think it makes sense to keep it as is. If they decide to change it once we get testers in it with the full flow we can change it if needed

apps/mobile-wallet-connector/src/app/connect-wallet/scan-code/page.tsx
escobarjonatan: ahh i see, this makes sense. Going to bring this up to make this the label of the code input

apps/mobile-wallet-connector/src/app/connect-wallet/scan-code/page.tsx
escobarjonatan: good catch, it doesn't. It did when i was looking at it before needing to redirect. The user will see this for .4 seconds and be confused. Removing.

scandycuz (approved):
LGTM

PR #595, opened 2024-04-18
[MRKT-34, MRKT-35]: adds more and similar songs to artist page

scandycuz:
Adds "More from {artist}" and "Similar songs" UI to item page.

### DIFF
--- a/apps/marketplace/src/app/item/[songId]/page.tsx
+++ b/apps/marketplace/src/app/item/[songId]/page.tsx
@@ -5,7 +5,7 @@ import { FunctionComponent, useEffect, useState } from "react";
 import { resizeCloudinaryImage } from "@newm-web/utils";
 import { ProfileImage } from "@newm-web/elements";
 import MoreSongs from "../../../components/MoreSongs";
-import { mockArtist, mockSongs } from "../../../temp/data";
+import { mockSongs } from "../../../temp/data";
 import { ItemSkeleton, SimilarSongs } from "../../../components";
 
 interface SingleSongProps {
@@ -17,7 +17,6 @@ interface SingleSongProps {
 const SingleSong: FunctionComponent<SingleSongProps> = ({ params }) => {
   const [isLoading, setIsLoading] = useState(true);
   const songData = mockSongs.find((song) => song.id === params.songId);
-  const artistData = mockArtist;
 
   // TEMP: simulate data loading
   useEffect(() => {
--- a/apps/marketplace/src/app/item/[songId]/page.tsx
+++ b/apps/marketplace/src/app/item/[songId]/page.tsx
@@ -91,9 +91,9 @@ const SingleSong: FunctionComponent<SingleSongProps> = ({ params }) => {
         </Stack>
       </Container>
 
-      <MoreSongs artist={ artistData } />
+      <MoreSongs />
 
-      <SimilarSongs song={ songData } />
+      <SimilarSongs />
     </Container>
   );
 };
--- a/apps/marketplace/src/app/item/[songId]/page.tsx
+++ b/apps/marketplace/src/app/item/[songId]/page.tsx
@@ -17,7 +17,7 @@ interface SingleSongProps {
 const SingleSong: FunctionComponent<SingleSongProps> = ({ params }) => {
   const [isLoading, setIsLoading] = useState(true);
   const songData = mockSongs.find((song) => song.id === params.songId);
-  const artist = mockArtist;
+  const artistData = mockArtist;
 
   // TEMP: simulate data loading
   useEffect(() => {
@@ -91,7 +91,7 @@ const SingleSong: FunctionComponent<SingleSongProps> = ({ params }) => {
         </Stack>
       </Container>
 
-      <MoreSongs artist={ mockArtist } />
+      <MoreSongs artist={ artistData } />
 
       <SimilarSongs song={ songData } />
     </Container>
--- a/apps/marketplace/src/app/item/[songId]/page.tsx
+++ b/apps/marketplace/src/app/item/[songId]/page.tsx
@@ -4,8 +4,9 @@ import { SongCard } from "@newm-web/components";
 import { FunctionComponent, useEffect, useState } from "react";
 import { resizeCloudinaryImage } from "@newm-web/utils";
 import { ProfileImage } from "@newm-web/elements";
-import { mockSongs } from "../../../temp/data";
-import { ItemSkeleton } from "../../../components";
+import MoreSongs from "../../../components/MoreSongs";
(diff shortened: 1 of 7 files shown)


apps/marketplace/src/app/item/[songId]/page.tsx
```diff
@@ -29,63 +30,69 @@ const SingleSong: FunctionComponent<SingleSongProps> = ({ params }) => {
...
+          </Box>
           <Stack
-            alignItems={ "center" }
-            direction={ "row" }
-            gap={ 1.5 }
-            justifyContent={ ["center", "center", "start"] }
+            gap={ [4, 4, 2.5] }
+            pt={ [0, 0, 1.5] }
+            textAlign={ ["center", "center", "left"] }
+            width={ ["100%", 440, 440] }
```
Comment (escobarjonatan) [inline apps/marketplace/src/app/item/[songId]/page.tsx:57]: really minor, but would a maxWidth be better here to have the content extend to 440 but not always be 440 if needed?

apps/marketplace/src/app/item/[songId]/page.tsx
```diff
@@ -29,63 +30,69 @@ const SingleSong: FunctionComponent<SingleSongProps> = ({ params }) => {
...
+              alignItems={ "center" }
+              direction={ "row" }
+              gap={ 1.5 }
+              justifyContent={ ["center", "center", "start"] }
+            >
+              <ProfileImage
+                height={ 40 }
+                src={ resizeCloudinaryImage(songData?.artist.profileImageUrl, {
+                  height: 56,
+                  width: 56,
```
Comment (escobarjonatan) [inline apps/marketplace/src/app/item/[songId]/page.tsx:81]: the height and the width for the ProfileImage are set to 40 but the height requested to resize is 56, is that ok?

apps/marketplace/src/temp/data.ts
```diff
@@ -19,6 +19,7 @@ export const mockArtist = {
   songCount: 4,
   soundCloudUrl: "https://www.example.com",
   spotifyUrl: "https://www.example.com",
+  stageName: "Dr. Shrimp Puerto Rico",
```
escobarjonatan, line 22: let's not put personal information in a public repo now 🫣

escobarjonatan (approved):

apps/marketplace/src/app/item/[songId]/page.tsx
```diff
@@ -29,63 +30,69 @@ const SingleSong: FunctionComponent<SingleSongProps> = ({ params }) => {
...
-            // eslint-disable-next-line @typescript-eslint/no-empty-function
-            onCardClick={ () => {} }
-            // eslint-disable-next-line @typescript-eslint/no-empty-function
-            onPriceClick={ () => {} }
-            // eslint-disable-next-line @typescript-eslint/no-empty-function
-            onSubtitleClick={ () => {} }
-          />
-        </Box>
+    <Container>
+      <Container maxWidth="md" sx={ { flexGrow: 1, mt: [0, 0, 5] } }>
```
Comment (dmkirshon) [inline apps/marketplace/src/app/item/[songId]/page.tsx:34]: A container within a container seems stranger than the other wrapper components, Stacks on Stacks, should we migrate the single song card/description/metadata into its own component? I can refactor this as part of the buy function/UI I'm working on if you agree.

apps/marketplace/src/app/item/[songId]/page.tsx
```diff
@@ -29,63 +30,69 @@ const SingleSong: FunctionComponent<SingleSongProps> = ({ params }) => {
...
+              </Typography>
+            </Stack>
           </Stack>
         </Stack>
-      </Stack>
+      </Container>
+
+      <MoreSongs />
+
+      <SimilarSongs />
```
Comment (dmkirshon) [inline apps/marketplace/src/app/item/[songId]/page.tsx:95]: There are skeletons for the individual songs as they load in but should we also have a larger block skeleton in `ItemSkeleton.tsx` to load with the top half of the Single Item/Song data?

apps/marketplace/src/components/MoreSongs.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const artist = mockArtist;
+  const songs = mockSongs;
+  const artistFullName = `${artist.firstName} ${artist.lastName}`;
+
+  const title = (
+    <Box component="span">
+      MORE FROM&nbsp;
+      <Box component="span" sx={ { color: theme.colors.music } }>
+        { artist.stageName || artistFullName }
+      </Box>
```
dmkirshon, line 18: The title doesn't seem to properly center on a mobile view like the "Similar Songs" title.

apps/marketplace/src/app/item/[songId]/page.tsx
Comment (dmkirshon) [reply apps/marketplace/src/app/item/[songId]/page.tsx:81]: From what Trevor explained to me it was due to the image looking not as sharp/a little pixelated when having the exact dimensions, not sure if we have a perfect ratio for this or not?

dmkirshon (approved):
LGTM, just a few minor questions/possible corrections.

apps/marketplace/src/app/item/[songId]/page.tsx
Comment (scandycuz) [reply apps/marketplace/src/app/item/[songId]/page.tsx:57]: Yeah, having a max width would be cleaner. Most of this code is actually unchanged, I only wrapped it in a container and added the `MoreSongs` and `SimilarSongs` components below. I imagine that could be a good change to make in a future refactor.

apps/marketplace/src/app/item/[songId]/page.tsx
Comment (scandycuz) [reply apps/marketplace/src/app/item/[songId]/page.tsx:81]: Yeah, I find that the images look a bit blurry unless the actual size is a bit larger than the rendered size. It may have something to do with the quality being lowered when Cloudinary resizes the image.

apps/marketplace/src/temp/data.ts
scandycuz: I just took it from the mocks, lol

apps/marketplace/src/app/item/[songId]/page.tsx
Comment (scandycuz) [reply apps/marketplace/src/app/item/[songId]/page.tsx:34]: The  MUI `Container` component [docs](https://mui.com/material-ui/react-container/) state that it's fine to nest containers. It works well here because both the top and bottom sections of the page are centered and the top section has a smaller max width.

Yeah, moving the top and details section of the page into separate components would be good.

apps/marketplace/src/app/item/[songId]/page.tsx
Comment (scandycuz) [reply apps/marketplace/src/app/item/[songId]/page.tsx:95]: Yeah, it would be good to move the top section of the page to it's own component and include the `ItemSkeleton` as part of that new component.

apps/marketplace/src/components/MoreSongs.tsx
scandycuz: Good catch, I'll push an update to center it.

PR #599, opened 2024-04-22
feat(aws): Test workflow for deploying to AWS

jayvaznewm:
This pull request demonstrates how to deploy web applications to AWS with live previews on each pull request.  If approved this same pattern will be applied to merging and deploying to studio environment.

studio:
-----
A static website and hosted on S3.  Live previews are available at https://{PR#}.[artist.preview.newm.io](http://artist.preview.newm.io/)

marketplace:
-----
marketplace is built as a next.js dynamic website with server side rendering and runs in a Docker container hosted on AWS Lambdas.  Hosting via a Docker container + Lamdba is an easy and flexible to start hosting with the option to expand to less ephemeral hosting in the future (ECS or EC2).  The preview URL will be posted to the PR when the deployment is finished.


escobarjonatan (approved):

.github/workflows/cdk/jest.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  testEnvironment: 'node',
+  roots: ['<rootDir>/test'],
+  testMatch: ['**/*.test.ts'],
+  transform: {
+    '^.+\\.tsx?$': 'ts-jest'
+  }
+};
```
scandycuz: Any way to reference the `jest.config` at the top level of the repo, or does cdk require it's own file? Also, do you know how this file is used when deploying? Should it match the jest.config used locally?

.vscode/settings.json
```diff
@@ -10,6 +10,6 @@
     "editor.defaultFormatter": "rvest.vs-code-prettier-eslint"
   },
   "[typescript]": {
-    "editor.defaultFormatter": "rvest.vs-code-prettier-eslint"
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
scandycuz: Was this intentional or required? We're currently using a formatter that combines prettier and eslint which seems to work well and is a bit more configurable.

apps/marketplace/project.json
```diff
@@ -29,11 +31,12 @@
         "development": {
           "buildTarget": "marketplace:build:development",
           "dev": true,
-          "port": 4200
+          "port": 3000
```
scandycuz: It would be good to leave the development port on 4200, so that each app can be run on a different port locally (although the back-end still needs CORS configured to support port 4200). The production port works as 3000 though.

scandycuz (changes requested):
Thanks again for putting this together. Not sure if any of my changes require holding off merging the PR, but wanted to leave a review with a status of "Request changes", to hold off merging in case some of them would be good to make. I imagine the PR may just be for testing though.

.github/workflows/cdk/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "aws-cdk": "2.140.0",
+    "ts-node": "^10.9.2",
+    "typescript": "~5.4.5"
+  },
+  "dependencies": {
+    "aws-cdk-lib": "2.140.0",
+    "constructs": "^10.0.0",
+    "source-map-support": "^0.5.21"
+  }
+}
```
scandycuz, line 23: Is it necessary to have a separate `package.json` file here? Nx seems to support having a single package.json file at the root of the repo, although I imagine cdk may require a separate one.

apps/marketplace/Dockerfile
```diff
@@ -0,0 +1,77 @@
+# Use the official Node.js 18 image as the base  
+FROM node:20 as base
```
scandycuz, line 2: I know Jonatan, David, and I are currently using Node v18, but we could upgrade to 20 if it's recommended.

.github/workflows/cdk/package.json
jayvaznewm: I am treating this cdk directory as a is a "mini-tool" only used to deploy into AWS - it coincidentally happens to be written in TypeScript.  Its small and single purpose, and I am not sure if it makes sense to integrate with the tooling of the actual project - so I kept the package.json separate - open to feedback tho.  Hopefully this CDK directory will be rarely touched

apps/marketplace/Dockerfile
jayvaznewm: Given the rate of new nodejs versions and the backward compatibility I tend to go with the latest stable version - both work here - so no concerns with 18 either

.github/workflows/cdk/jest.config.js
jayvaznewm: since the cdk project is small and single purpose - its unlikely we'll ever develop tests for it - so I just nuked it

.vscode/settings.json
jayvaznewm: sorry - copy pasta - fixed now

.github/workflows/cdk/jest.config.js
scandycuz: Sounds good. I mostly was just unsure if it was going to be running the app unit tests with that jest configuration, but it sounds like it would be for separate cdk related tests.

.github/workflows/cdk/package.json
scandycuz: Having it separate sounds like it makes sense

apps/marketplace/Dockerfile
scandycuz: Either way works for me. It looks like the default version on the Node.js website is currently 20, so upgrading is probably a good idea

scandycuz (approved):
LGTM!

PR #605, opened 2024-04-29
MRKT-57: Update song card design

escobarjonatan:
image

### DIFF
--- a/apps/studio/src/pages/home/owners/Songs.tsx
+++ b/apps/studio/src/pages/home/owners/Songs.tsx
@@ -166,7 +166,6 @@ const Songs: FunctionComponent = () => {
               isPlaying={ song.id === playerState.currentPlayingSongId }
               subtitle={ genresString }
               title={ song.title }
-              onCardClick={ () => handleSongPlayPause(song) }
               onPlayPauseClick={ () => handleSongPlayPause(song) }
             />
           </Grid>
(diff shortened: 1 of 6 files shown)


apps/studio/src/pages/home/owners/Songs.tsx
```diff
@@ -166,6 +166,7 @@ const Songs: FunctionComponent = () => {
               isPlaying={ song.id === playerState.currentPlayingSongId }
               subtitle={ genresString }
               title={ song.title }
+              onCardClick={ () => handleSongPlayPause(song) }
```
escobarjonatan: Currently in Studio, the card has "clickable" features(hover,background,pointer) so i added the handler to be the same as play/pause. The other option was to remove the clickable features if the handler wasn't passed(so i would remove this line)

packages/components/src/lib/SongCard.tsx
```diff
@@ -93,16 +77,19 @@ export const SongCard = ({
     );
   }
 
+  const CardComponent = onCardClick ? Clickable : Box;
```
dmkirshon: `Clickable` already seems to handle the scenario when the `onClick` is fed `undefined` through your added ternaries below, not sure if `CardComponent` is needed. Then padding in `sx` seems to be also redundant. Is this to keep it more explicit between Studio and Marketplace?

dmkirshon (approved):
LGTM just a comment on the Component conditional in `SongCard.tsx`

scandycuz (approved):
LGTM

packages/components/src/lib/SongCard.tsx
escobarjonatan: Yes you are 100% about the onclick, i started off with 1 approach to remove the handler but forgot to put it back together lol good catch. The 1.5 padding was needed with the Box implementation but since we dont need that i removed it

PR #606, opened 2024-04-30
STUD-216: adds minimum file size for song image upload

scandycuz:


packages/elements/src/lib/UploadImage.tsx
dmkirshon, line 1: Noticed an eslint warning for `React Hook useCallback has a missing dependency: 'minFileSizeMB'.`

dmkirshon (approved):
LGTM just noticed an `eslint[react-hooks/exhaustive-deps]` warning in `UploadImage.tsx`

apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
```diff
@@ -282,6 +282,7 @@ const BasicSongDetails: FunctionComponent<BasicDonDetailsProps> = ({
               emptyMessage="Drag and drop or browse your image"
               maxFileSizeMB={ 10 }
               minDimensions={ { height: 1400, width: 1400 } }
+              minFileSizeMB={ 0.1 }
```
AndrewWestberg, line 285: @wlara If we store a cloudinary URL in our database and the user has uploaded a 100kb png that's 1600x1600, won't cloudinary serve us a resized 1400x1400 that could be less than 100kb?

packages/elements/src/lib/UploadImage.tsx
scandycuz: Good catch, added the dependency.

apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
scandycuz: @AndrewWestberg My understanding is that Cloudinary only downsizes the image if it's larger than 4,000 x 4,000 pixels, in which case it would be downsized to 4,000 x 4,000. I'm not familiar with any resizing that's done on the back-end though.

PR #608, opened 2024-05-02
MRKT-24: sale data

scandycuz:


scandycuz:
Marked as a draft as there's still a few more back-end changes that need to be made.

apps/mobile-wallet-connector/src/api/utils.ts
```diff
@@ -31,59 +30,3 @@ export const prepareHeaders = async (
     ...headers,
   };
 };
```
scandycuz, line 32: Moved `axiosBaseQuery` to the `utils` package, since it should be the same for all apps.

packages/utils/src/lib/audio.ts
```diff
@@ -0,0 +1,69 @@
...
+        },
+        onplay: (id) => {
+          setAudioUrl(src);
+          setIsAudioPlaying(true);
+        },
+        onstop: () => {
+          setIsAudioPlaying(false);
+        },
+        src,
+      });
```
dmkirshon, line 59: Just a heads up, ran into issues with playback in my progress bar branch,  if the source is going to be the same as the nft Asset previews/clips like pool.pm through nftcdn, then we will need to add a `format` prop as they don't have filetype endings, which howl doesn't appreciate. I believe we only support flac and wav, not sure if the clip song preview will be in a different format.

`format: ["wav", "flac"],`

[Source of Information](https://www.npmjs.com/package/howler#src-arraystring--required:~:text=If%20your%20files%20have%20no%20extensions%2C%20you%20will%20need%20to%20explicitly%20specify%20the%20extension%20using%20the%20format%20property.)

scandycuz: Good call-out. Once the API data is updated with the playable audio files, I'll make sure that everything is playing correctly.

scandycuz: I was running into the issue as well, but having `html: true` looks to make it work.

apps/marketplace/project.json
```diff
@@ -29,7 +29,7 @@
         "development": {
           "buildTarget": "marketplace:build:development",
           "dev": true,
-          "port": 4200
+          "port": 3000
```
escobarjonatan: We should coordinate with James instead of committing this here that way we can run multiple apps at the same time (i do this when working on common code)

apps/marketplace/src/app/sale/[saleId]/page.tsx
```diff
@@ -8,37 +8,82 @@ import {
...
+  const router = useRouter();
 
-  // TEMP: simulate data loading
-  useEffect(() => {
-    setTimeout(() => {
-      setIsLoading(false);
-    }, 1000);
-  }, []);
+  const { isAudioPlaying, playPauseAudio } = usePlayAudioUrl();
+  const { isLoading, data: sale } = useGetSaleQuery(params.saleId);
```
Comment (escobarjonatan) [inline apps/marketplace/src/app/sale/[saleId]/page.tsx:39]: I am thinking we should probably handle the case where sale is empty? or if the `isError` from this query returns true?

apps/marketplace/src/app/sale/[saleId]/page.tsx
```diff
@@ -8,37 +8,82 @@ import {
...
+      newmAmount: formatNewmAmount(newmAmount),
+      usdAmount: currency(usdAmount).format(),
+    };
+  };
+
+  /**
+   * Navigates to the artist page when clicked.
+   */
+  const handleArtistClick = () => {
+    if (!sale) return;
```
Comment (escobarjonatan) [inline apps/marketplace/src/app/sale/[saleId]/page.tsx]: handling the `!sale` before all of this logic gets executed might help us not have to account for this case in multiple handlers

apps/marketplace/src/modules/sale/types.ts
```diff
@@ -0,0 +1,117 @@
...
+  Ended = "Ended",
+  SoldOut = "SoldOut",
+  Started = "Started",
+}
+
+export interface SaleState {
+  readonly sales: ReadonlyArray<Sale>;
+}
+
+export interface Sale {
```
escobarjonatan, line 11: Are any of these types map 1:1 with endpoint documentation? If so, might be worth calling that out somewhere so next time we need to update it we know it's ok to do so to maintain the consistent 1:1 map between FE and BE

apps/marketplace/src/modules/sale/types.ts
```diff
@@ -0,0 +1,117 @@
...
+export interface Song {
+  // UUID of the song artist
+  readonly artistId: string;
+  // Stage name of the song artist
+  readonly artistName: string;
+  // url for the song artist's profile image
+  readonly artistPictureUrl: string;
+  // Valid URL of song audio clip file
+  readonly clipUrl: string;
+  // Song collaborator objects (see details bellow).	Always
```
escobarjonatan: NIT: Too many spaces?
`// Song collaborator objects (see details bellow).	Always`

apps/marketplace/src/modules/sale/utils.ts
```diff
@@ -0,0 +1,21 @@
+import { ApiSale, Sale } from "./types";
+
+/**
+ * Creates a sale object from the sale API object with
+ * the song.parentalAdvisory string field replaced with
+ * a song.isExplicit boolean field.
+ */
+export const transformApiSale = (apiSale: ApiSale): Sale => {
```
escobarjonatan, line 8: 🔥

escobarjonatan (approved):

apps/marketplace/project.json
scandycuz: I'll revert the change.

apps/marketplace/src/app/sale/[saleId]/page.tsx
Comment (scandycuz) [reply apps/marketplace/src/app/sale/[saleId]/page.tsx:39]: Yeah, we should reach out to design about some sort of error state UI when it fails.

apps/marketplace/src/modules/sale/types.ts
scandycuz: Yeah, all of the fields are what the API returns, except for the `isExplicit` field, which replaces the `parentalAdvisory` field in the endpoint `tranformResponse` functions. I'm not sure we need to call it out, as the `GetSaleResponse` interface denotes that it is the same as the `Sale` interface, and the `ApiSale` interface denotes which field differs in the API data.

I'm open to adding some comments to make it more obvious though.

apps/marketplace/src/modules/sale/types.ts
scandycuz: That was copied and pasted from the API docs, and it looks like the space is  display strangely in Github. The "Always" was meant to be removed.

apps/marketplace/src/app/sale/[saleId]/page.tsx
Comment (scandycuz) [reply apps/marketplace/src/app/sale/[saleId]/page.tsx]: That could work, although it is nice to have all of the return functions at the bottom of the component. For now, I'll just replace the early return with a thrown error, as it shouldn't actually ever occur if we have the error UI state.

PR #609, opened 2024-05-03
STUD-223: Add country to signup flow and make it editable in profile

escobarjonatan:
https://github.com/projectNEWM/newm-web/assets/11530120/48f6a8ae-190a-4949-9cae-efd8c4c019e6

### DIFF
--- a/apps/studio/src/common/formUtils.ts
+++ b/apps/studio/src/common/formUtils.ts
@@ -159,12 +159,7 @@ export const commonYupValidation = {
     .trim()
     .max(20, "Must be 20 characters or less")
     .required("Last name is required"),
-  location: (locations: string[]) =>
-    Yup.string()
-      .required("This field is required")
-      .test("is-location", "You need to type or select one", (location) =>
-        locations.includes(location || "")
-      ),
+  location: Yup.string().required("This field is required"),
   moods: Yup.array().max(5, "Maximum of 5 moods allowed"),
   newPassword: Yup.string()
     .test(
--- a/apps/studio/src/common/formUtils.ts
+++ b/apps/studio/src/common/formUtils.ts
@@ -159,6 +159,12 @@ export const commonYupValidation = {
     .trim()
     .max(20, "Must be 20 characters or less")
     .required("Last name is required"),
+  location: (locations: string[]) =>
+    Yup.string()
+      .required("This field is required")
+      .test("is-location", "You need to type or select one", (location) =>
+        locations.includes(location || "")
+      ),
   moods: Yup.array().max(5, "Maximum of 5 moods allowed"),
   newPassword: Yup.string()
     .test(
--- a/apps/studio/src/pages/createProfile/CreateProfile.tsx
+++ b/apps/studio/src/pages/createProfile/CreateProfile.tsx
@@ -9,7 +9,7 @@ import Complete from "./Complete";
 import AddFirstName from "./AddFirstName";
 import AddLastName from "./AddLastName";
 import SelectLocation from "./SelectLocation";
-import { useGetCountriesQuery, useGetRolesQuery } from "../../modules/content";
+import { useGetRolesQuery } from "../../modules/content";
 import { commonYupValidation } from "../../common";
 import {
   ProfileFormValues,
@@ -19,7 +19,6 @@ import {
 const CreateProfile: FunctionComponent = () => {
   const theme = useTheme();
   const { data: roles = [] } = useGetRolesQuery();
-  const { data: locations = [] } = useGetCountriesQuery();
 
   const [updateInitialProfile] = useUpdateInitialProfileThunk();
 
@@ -39,7 +38,7 @@ const CreateProfile: FunctionComponent = () => {
   const validations = {
     firstName: commonYupValidation.firstName,
     lastName: commonYupValidation.lastName,
-    location: commonYupValidation.location(locations),
+    location: commonYupValidation.location,
     nickname: commonYupValidation.nickname,
     role: commonYupValidation.role(roles),
   };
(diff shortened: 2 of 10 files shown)


packages/elements/src/lib/styled/NoResultsText.ts
```diff
@@ -17,8 +17,10 @@ const NoResultsText = styled("span")`
   padding: 12px 12px;
   position: absolute;
   width: 100%;
-  z-index: 1;
-
+  z-index: 10;
+  bottom: -52px;
```
escobarjonatan, line 21: For some strange reason, i am not sure why this hadn't been an issue before, but we are placing this absolute to the relative field(dropdown) but were not setting the place of where we wanted it. It worked in the scenarios we are using it in currently assuming bc of the containers we are wrapping it on? anyway, i tried to use the dropdown field as it was and the results came out fine but the no results box was somewhere else lol I tested out the other places to make sure this still worked as we intended and it does. The 52 offsets the input

apps/studio/src/pages/createProfile/CreateProfile.tsx
```diff
@@ -96,6 +100,13 @@ const CreateProfile: FunctionComponent = () => {
                 role: validations.role,
               }),
             },
+            {
+              element: <SelectLocation />,
+              path: "what-is-your-loocation",
```
dmkirshon: Was wondering if I was having a stroke... just looks like one too many o's in loooocation. Was sending me to a black page with invitations pending as the `!location` redirect is correct https://github.com/projectNEWM/newm-web/blob/17f95e91fcf7cbee7f759a691325e81f804cf26b/apps/studio/src/pages/home/Home.tsx#L35-L38


image

apps/studio/src/pages/createProfile/SelectLocation.tsx
```diff
@@ -0,0 +1,102 @@
...
+    <Box alignItems="center" display="flex" flexDirection="column">
+      <Box mb={ 4 }>
+        <ResponsiveNEWMLogo />
+      </Box>
+      <Typography
+        id="location-prompt"
+        sx={ { textAlign: "center" } }
+        variant="h1"
+      >
+        Finally, what is your country of residence?
```
dmkirshon, line 47: The UX seems ok for our initial release but if we end up having more required details for them to fill out we might want to think of a different UX as this seems pretty jarring to just redirect to this page. It might just be the word "Finally" for existing users that's catching me off guard...

apps/studio/src/common/formUtils.ts
```diff
@@ -159,6 +159,12 @@ export const commonYupValidation = {
     .trim()
     .max(20, "Must be 20 characters or less")
     .required("Last name is required"),
+  location: (locations: string[]) =>
+    Yup.string()
+      .required("This field is required")
+      .test("is-location", "You need to type or select one", (location) =>
```
dmkirshon: Will this test ever be reached? We give them set location values and they have to select one. Trying to manually type something in the field just clears out if it doesn't select a value.

apps/studio/src/pages/createProfile/CreateProfile.tsx
dmkirshon, line 1: I don't think I have a `location` set in my Profile. I keep getting redirected to the `create-profile/location` page, it seems like my location value in the create-profile flow doesn't ever get updated so I'm stuck in an endless loop. I haven't been able to pinpoint why my selected `location` value doesn't persist. `updateInitialProfile` is called but it doesn't include the `location` value in the update. 

When I select next and move from the `create-profile/location` page to `create-profile/complete` it looks like formik loses `FormikContext: {values:  }`. It persists when starting on the initial `create-profile` route, but starting on any of the other routes leads into the same issue where the next button clears `FormikContext: {values}`. This might be a bug within `WizardForm` and how things persist when entering the form midway? 

I need to do a bit more digging if you don't see this on your end. If you already have a saved `location` in your Profile, have you chosen a different country on the `create-profile/location` page from the one that is in your Profile and seen it persist?

dmkirshon (changes requested):
Running into an issue with form values not persisting on the `create-profile` flow causing an endless user loop. Also a few other comments and corrections noted.

apps/studio/src/pages/createProfile/CreateProfile.tsx
escobarjonatan: i knew i should have copy/paste lol good catch, adding 3 more o's

apps/studio/src/pages/createProfile/SelectLocation.tsx
escobarjonatan: This will be the experience for a few of them, most will go through the signup flow. This should be ok

apps/studio/src/common/formUtils.ts
escobarjonatan: i copied it from roles, but you are right, it doesn't seem to get to it, i'll remove it

apps/studio/src/pages/createProfile/CreateProfile.tsx
escobarjonatan: Very nice find, see i already had my location setup before this and tried to mock it without it but clearly didnt work lol I found out the issue(i think). In your case, you didn't have your location but you had your name/last/role and other stuff. so it went like this:
1. location page -> 2. all set (submit) -> 3. profile page -> 4. redirect to create-profile page


on 1, the form had the initial values to empty strings(name/last/role) so when submitting your location, you were also submitting empty values for the rest, meaning that you were setting your name to "" and that's where the loop began. I fixed this by changing the initial values to the profile query if it had any of those values it would use them as the initial values. Here's a video to demo this flow. I temporarily made the input on Profile a text and removed validation for this demo only: 

https://github.com/projectNEWM/newm-web/assets/11530120/948c2d96-df65-4b37-a3f8-2a617c3218ba

dmkirshon (approved):
LGTM, looks like the flow is working now!

scandycuz (approved):

PR #617, opened 2024-05-13
Add Progress Bar to Song Card (Studio & Marketplace)

dmkirshon:
Add:
- Progress bar to track audio playback on sound card component

Note: Change affects all sound cards
1) Sale page on Marketplace
2) All Sales previews (Just Released, More From, Similar Songs) on Marketplace
3) Collaborators -> Artist Page with collab previews on Studio

https://github.com/projectNEWM/newm-web/assets/86050631/a95a4762-26f0-498d-99e4-c9e55f6504d4

https://github.com/projectNEWM/newm-web/assets/86050631/1048ee63-f0e2-4102-b677-9a20921b7729


packages/components/src/lib/SongCard.tsx
```diff
@@ -38,10 +53,33 @@ const SongCard = ({
...
+  title,
 }: SongCardProps) => {
   const theme = useTheme();
 
+  const [songProgress, setSongProgress] = useState(0);
+
+  useEffect(() => {
+    if (isPlaying && duration) {
+      // Gives indication that the song is playing by incrementing the progress bar
+      if (songProgress === 0) setSongProgress(1);
```
dmkirshon: I added this to show progress of the playback from the start of track audio, but no issue removing it too. I think we need to add an extra one second to the duration because of this, but would like to see your and product's opinion.

Standard across other platforms: Would need to remove the line, the first second of progress is typically delayed by the interval loop. Seen on Spotify and others.

apps/marketplace/src/temp/data.ts
```diff
@@ -51,15 +51,15 @@ export const mockSales: ReadonlyArray<Sale> = [
     costAssetName: "XYZ123",
     costPolicyId: "123XYZ",
     createdAt: new Date("April 10th, 2024").toDateString(),
-    id: "3cfb2d02-a320-4385-96d1-1498d8a1df58",
+    id: "cec4f704-69c4-41fd-807d-47aa2d73f0d2",
```
dmkirshon, line 54: Added this just for testing, this is one of the backend IDs, so clicking the first track under `Just Released` leads to a backend example. The First two play icons under `Just Released` can playback from a `nftcdn` source that might be similar to what the end url will look like.

Didn't want to change this file up too much since we might be pulling in backend data soon and replacing all of this.

apps/studio/src/pages/home/owners/Songs.tsx
dmkirshon, line 1: Noticed a long delay (5 seconds) before the Studio sound card is actually playing audio on initial click. The progress bar is rolling from the initial click... need to figure out how to sync up with the network loading delay. Will create a separate bug ticket for this, probably need to expose the actual audio position which will then help with adding a seek functionality.

apps/marketplace/src/components/Sales.tsx
dmkirshon, line 1: The Sales component and Sale page are separated so the playback of the sound card doesn't know the other instance exists. Aka you can play the Sale page asset and one of the More of Artist SongCards at the same time. 

Will create another ticket/branch to include Redux or universal playback tracking across marketplace.

scandycuz: Sounds good. It may have already been your intention, but I'd also incorporate the Redux functionality into the hooks so that it's abstracted there.

escobarjonatan (approved):

scandycuz (changes requested):
Looking good, but I'd recommend getting the actual song position from the song itself,  that should resolve some of the issues around how to accurately display the song position.

apps/studio/src/pages/home/owners/Songs.tsx
scandycuz: I'd recommend refactoring the code to get the position from the song as part of this ticket, since it would be good for the progress bar to reflect the position of the song for the MVP. The seek functionality doesn't need to be in the MVP though.

apps/marketplace/src/modules/sale/api.ts
```diff
@@ -71,7 +71,7 @@ export const extendedApi = newmApi.injectEndpoints({
           ...(statuses ? { statuses: statuses.join(",") } : {}),
           ...rest,
         },
-        url: "/v1/marketplace/sales",
+        url: "v1/marketplace/sales",
```
escobarjonatan, line 74: good catch, i am bad with these 😅

escobarjonatan (approved):

packages/utils/src/lib/hooks.ts
```diff
@@ -211,9 +238,9 @@ export const useHlsJs = ({
       }
 
       handlePlaySong(song);
-
       videoRef.current.addEventListener("ended", handleSongEnded);
     },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
scandycuz: This works, but in order to not have to ignore the line, it should be possible to create all dependent functions using `useCallback`, if the concern is that including the functions as dependencies will cause the function to be called more often than it should. I didn't test it, so I imagine the comment may be necessary if that approach still results in the function being called at the incorrect time, but wanted to mention it in case.

Also, if it is necessary, it can be helpful to add another comment above the `eslint-disable-next-line` comment, just to clarify why including all dependencies doesn't work.

scandycuz (approved):
LGTM

packages/utils/src/lib/hooks.ts
dmkirshon: Added the callbacks, and added the appropriate `eslint-disable` with the functions, `playSongWithHlsJs`  and `playSongNatively`, that would have caused issues if dependencies of `trackSongProgress` were included.

PR #627, opened 2024-05-17
[MRKT-34, MRKT-35]: fetch API data for sale songs by artist and similar songs

scandycuz:


apps/marketplace/src/components/Sale.tsx
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    router.push(`/artist/${artistId}`);
+  };
+
+  if (isLoading) {
+    return <ItemSkeleton />;
+  }
+
+  if (!sale) return null;
```
escobarjonatan, line 85: Because of this condition in line 87, i dont think the other conditions in lines 49, 59 & 76 will ever execute and therefore not needed 🤷🏽‍♂️

escobarjonatan (approved):

apps/marketplace/src/components/Sale.tsx
scandycuz: True, the functions would never actually be called without a sale, but technically it's still undefined in the code. We could move the early return near the top of the function, but I think it's more readable to have all of the return statements at the bottom. Removed it from `handleArtistClick` though.

PR #632, opened 2024-05-19
MRKT-14: search sales by typing

scandycuz:


packages/components/src/lib/SongCard.tsx
```diff
@@ -43,21 +43,27 @@ const SongCard = ({
...
+    onPlayPauseClick();
   };
 
   const handleSubtitleClick = (event: MouseEvent | KeyboardEvent) => {
+    if (!onSubtitleClick) return;
+
     event.preventDefault();
     event.stopPropagation();
-    onSubtitleClick?.();
+    onSubtitleClick();
```
scandycuz, line 66: Updated these so that clicking the element doesn't prevent the element beneath it from being clicked when handler isn't present.

apps/marketplace/src/components/Sales.tsx
```diff
@@ -40,23 +42,31 @@ const Sales: FunctionComponent<SalesProps> = ({
...
-            <Typography sx={ { marginTop: 8, textAlign: "center" } }>
-              No songs to display at this time.
-            </Typography>
+            { typeof noResultsContent === "string" ? (
+              <Typography sx={ { marginTop: 8, textAlign: "center" } }>
+                { noResultsContent }
+              </Typography>
+            ) : (
+              noResultsContent
+            ) }
```
dmkirshon, line 69: Is this just prep for some other content? Is design/product working on something to place here instead of the text?

dmkirshon (approved):
LGTM

apps/marketplace/src/app/search/page.tsx
```diff
@@ -1,50 +1,21 @@
...
-          // eslint-disable-next-line @typescript-eslint/no-empty-function
-          onCardClick={ () => {} }
-          // eslint-disable-next-line @typescript-eslint/no-empty-function
-          onSubtitleClick={ () => {} }
-        />
-      </Stack>
+      <SearchResults query={ searchTerm } />
     </Container>
   );
 };
```
escobarjonatan, line 15: I think we should consider this if `!searchTerm` is present. 
```
"use client";
import { Container } from "@mui/material";
import { useRouter, useSearchParams } from "next/navigation";
import { SearchResults } from "../../components";
import { useAppDispatch } from "../../common";
import { setToastMessage } from "../../modules/ui";

const Search = () => {
  const searchParams = useSearchParams();
  const router = useRouter();
  const searchTerm = searchParams.get("searchTerm") || "";

  if (!searchTerm) {
    dispatch(
      setToastMessage({
        message:
          "No search term provided. Please enter a search term and try again.",
        severity: "error",
      })
    );

    router.push("/");
  }

  return (
    <Container sx={ { flexGrow: 1, mt: 5 } }>
      <SearchResults query={ searchTerm } />
    </Container>
  );
};

export default Search;
useSearchParams();
  const router = useRouter();
  const searchTerm = searchParams.get("searchTerm") || "";

  if (!searchTerm) {
    dispatch(
      setToastMessage({
        message:
          "There is a problem with your wallet. Please try disconnecting and reconnecting it.",
        severity: "error",
      })
    );

    router.push("/");
  }

  return (
    <Container sx={ { flexGrow: 1, mt: 5 } }>
      <SearchResults query={ searchTerm } />
    </Container>
  );
};

export default Search;

```

The current implementation has unreachable code: `return null;` (also gets called out by my IDE). It also leaves the app in an error state that the user might not know what to do? 

With the changes suggested:
- Display a toast message letting them know the error
- Redirect the user to the home page (that also has the search available)

apps/marketplace/src/components/Sales.tsx
```diff
@@ -75,7 +85,6 @@ const Sales: FunctionComponent<SalesProps> = ({
                   title={ song.title }
                   onCardClick={ () => handleCardClick(id) }
                   onPlayPauseClick={ () => playPauseAudio(song.clipUrl) }
-                  onSubtitleClick={ () => handleSubtitleClick(id) }
```
escobarjonatan: I see why you are removing it, however, instead of removing it we should change the subtitle from the current `genresString` to `subtitle={ song.artistName }` And on subtitle click it would take the user to the artist page. So these changes:

Replace:
`subtitle={ song.artistName }`
to
`subtitle={ genresString }` 

Replace:
`onSubtitleClick={ () => handleSubtitleClick(id) }`
to
`onSubtitleClick={ () => handleSubtitleClick(song.artistId) }`

apps/marketplace/src/app/search/page.tsx
```diff
@@ -1,50 +1,21 @@
 "use client";
```
escobarjonatan, line 1: There’s a weird bug in I noticed.

Steps:
1. In the search bar of the home page, type w  and and submit the search
2. Now add 1 space before the letter w, so this  w  and submit, the url should look something like this: http://localhost:3000/search/?searchTerm=%20w
3. Add 1 more space before the space and w so this:   w and submit, the url should now look like this: http://localhost:3000/search/?searchTerm=%20%20w
4. Notice how the UI is showing 1 item(screenshot), but the network call returned an empty array

image

apps/marketplace/src/components/Sales.tsx
scandycuz: There isn't a mock-up for when the search page doesn't return any results, but I implemented it so that it has custom text when no results are returned. It's also left aligned, so it's passed in as a React node with styling to make it left aligned instead of just a string, which would be center aligned by default.

Another way to do it could be to have a `noResultsContent` string prop, and a `noResultsContentAlignment` prop to set the text alignment, but it seemed easier to just allow passing a single React node prop.

apps/marketplace/src/app/search/page.tsx
scandycuz: It's strange the linter is throwing an error about retuning a null value, it shouldn't do that if the null value will only be returned in some cases. In retrospect, I'm thinking we can probably just return an empty result if the search term is empty or missing. The user shouldn't actually ever be able to see that page with an empty or missing search string, and if they do, by manually changing the url, then I think just displaying a message that no results were returned should work.

apps/marketplace/src/components/Sales.tsx
scandycuz: Ah, yeah. Seems the mock-ups were updated.

apps/marketplace/src/app/search/page.tsx
scandycuz: This looks to be because that song is a duplicate in the test database, which is causing issues with the React rendering because they have the same song ID, which is being used as the `key` prop. I updated the key to be the sale ID, which is unique and resolved the issue.

escobarjonatan: ahh, makes sense, i saw that key issue somewhere else in the console

escobarjonatan (approved):

PR #633, opened 2024-05-19
Misc fixes

scandycuz:
- Updates song card price layout to stack vertically.
- limits "More from {artist}" and "Similar songs" to 8 songs to not make it difficult to scroll to other sections. Post-MVP, some sort of "View all" button can be added.

### DIFF
--- a/packages/components/src/lib/skeletons/SongCardSkeleton.tsx
+++ b/packages/components/src/lib/skeletons/SongCardSkeleton.tsx
@@ -29,10 +29,15 @@ const SongCardSkeleton: FunctionComponent<SongCardSkeletonProps> = ({
     >
       <Skeleton height="100%" variant="rectangular" width="100%" />
       <Stack direction="row" justifyContent="space-between">
-        { isTitleVisible && <Skeleton height={ 30 } width={ 100 } /> }
-        { isPriceVisible && <Skeleton height={ 30 } width={ 80 } /> }
+        <Stack direction="column">
+          { isTitleVisible && <Skeleton height={ 30 } width={ 100 } /> }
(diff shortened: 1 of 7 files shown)


packages/components/src/lib/skeletons/SongCardSkeleton.tsx
dmkirshon, line 1: The `ItemSkeleton.tsx`, (which probably should be renamed to `SaleSkeleton.tsx`) also doesn't match up to the new price layout. 

https://github.com/projectNEWM/newm-web/blob/d53fb12660eae3627b94c333a32bdf6484cc98b9/apps/marketplace/src/components/skeletons/ItemSkeleton.tsx#L4

dmkirshon (approved):
LGTM, just minor comment

dmkirshon:
Noticed this through my review as well but not directly related, must have been a recent PR. Saw an error with the `Sale` type being imported for the `Sale` component. Minor conflict but probably should be either refactored with a renaming of the two or at least be updated to the following to remove the error:

```
import type { Sale } from "../modules/sale";
```

https://github.com/projectNEWM/newm-web/blob/d53fb12660eae3627b94c333a32bdf6484cc98b9/apps/marketplace/src/components/Sale.tsx#L19-L26

escobarjonatan (approved):

packages/components/src/lib/skeletons/SongCardSkeleton.tsx
scandycuz: It looks like the price should actually still appear in the upper right corner for the actual item. I'll make a separate ticket to make that update, since that requirement wasn't clear when changing the way it appeared for the previews.

scandycuz:
Resolved the import name conflict.

PR #656, opened 2024-05-26
MRTK-26: implements artist spotlight with API data

scandycuz:


apps/marketplace/src/components/ArtistSpotlight.tsx
```diff
@@ -1,61 +1,28 @@
...
-
-  // TEMP: simulate data loading
-  useEffect(() => {
-    setTimeout(() => {
-      setIsLoading(false);
-    }, 1000);
-  }, []);
+  const withActiveSales = artists.filter(({ marketplaceSongCount }) => {
+    return marketplaceSongCount > 0;
+  });
```
escobarjonatan: this could be a double edge sword lol What if none of the artists you pulled have active sales?

apps/marketplace/src/modules/artist/api.ts
```diff
@@ -0,0 +1,64 @@
...
+export const extendedApi = newmApi.injectEndpoints({
+  endpoints: (build) => ({
+    getArtist: build.query<GetArtistResponse, string>({
+      async onQueryStarted(body, { dispatch, queryFulfilled }) {
+        try {
+          await queryFulfilled;
+        } catch (error) {
+          dispatch(
+            setToastMessage({
+              message: "An error occurred while fetching artists",
```
escobarjonatan: >An error occurred while fetching the artist

Since it's a single lookup

escobarjonatan (approved):

apps/marketplace/src/components/ArtistSpotlight.tsx
scandycuz: Good point, this needs to be added as an API param so the list will return the first 8 with active sales

apps/marketplace/src/modules/artist/api.ts
scandycuz: Updated

PR #662, opened 2024-05-29
MRKT-78: similar artist section to API data

scandycuz:


apps/marketplace/src/components/SimilarArtists.tsx
```diff
@@ -1,54 +1,29 @@
...
-              );
-            }
-          ) }
-      </Grid>
+      <Artists
+        artists={ artists }
+        hasTitle={ true }
+        isLoading={ isLoading }
+        itemOrientation="row"
+        numSkeletons={ 6 }
```
escobarjonatan: We could have this 6 be a variable and we use it here and in the limit as well.

apps/marketplace/src/components/SimilarArtists.tsx
```diff
@@ -1,54 +1,29 @@
...
-    setTimeout(() => {
-      setIsLoading(false);
-    }, 1000);
-  }, []);
+const SimilarArtists: FunctionComponent<SimilarArtistsProps> = ({ genre }) => {
+  // TODO: limit to artists with marketplace sales when API updated
+  const { isLoading, data: artists = [] } = useGetArtistsQuery({
+    genres: genre ? [genre] : [],
+    limit: 6,
+  });
```
escobarjonatan: Do we need to add anything to exclude the current artist or will it automatically do that?

apps/marketplace/src/components/Artists.tsx
```diff
@@ -49,29 +50,37 @@ const Artists: FunctionComponent<ArtistsProps> = ({
...
+            return (
+              <Grid
+                display="flex"
+                justifyContent="center"
+                key={ id }
+                lg={ gridSizeColumnMap.lg[itemOrientation] }
+                md={ gridSizeColumnMap.md[itemOrientation] }
+                sm={ gridSizeColumnMap.sm[itemOrientation] }
+                xs={ gridSizeColumnMap.xs[itemOrientation] }
+                item
```
escobarjonatan: Is this a Boolean? If so, we can make it explicit item={true}

escobarjonatan (approved):

apps/marketplace/src/components/Artists.tsx
scandycuz: I'm ok with that. We should probably update the eslint config if we want to standardize on that.

apps/marketplace/src/components/SimilarArtists.tsx
scandycuz: Good point. It should be returning the current artist without that. I'll create a bug ticket.

apps/marketplace/src/components/MoreSongs.tsx
```diff
@@ -37,7 +42,7 @@ const MoreSongs: FunctionComponent<MoreSongsProps> = ({
 
   return (
     <Box mt={ [10, 8, 12.5] }>
-      <Sales isLoading={ isLoading } sales={ sales } title={ title } />
+      <Sales isLoading={ skip || isLoading } sales={ sales } title={ title } />
```
escobarjonatan, line 45: if `skip` is true, it means no data will be fetch and will be loading until user refreshes? might be worth handling that separate and show some text to action the user to refresh?

scandycuz: Without `isLoading={ skip || isLoading }`, the UI will show a "no results" state and then a loading state. This ensures that the loading state is present even when the params to fetch the data aren't present yet, and that no uncecessary requests are made before the params are present.

scandycuz: If there's some sort of error and the request is never actually made, the user should see a toast that a request failed. We could look into adding some some sort of error state in the UI in the future for that use case.

apps/marketplace/src/components/SimilarArtists.tsx
scandycuz: The bug's been resolved.

PR #669, opened 2024-05-30
Chore: Update urls and reenabled deployment in studio

escobarjonatan:


.github/workflows/aws-hosting-deploy-published-release-newm-studio.yml
```diff
@@ -1,9 +1,8 @@
 name: Deploy NEWM Studio app to AWS on published release
 on:
-#  release:
-#    types: [ published ]
-# only manually trigger for testing
-  workflow_dispatch:
+ release:
+   types: [ published ]
+ workflow_dispatch:
```
escobarjonatan, line 5: For some reason this was commented out. I believe we want to include this. @scandycuz can you confirm

scandycuz (approved):

.github/workflows/aws-hosting-deploy-published-release-newm-studio.yml
scandycuz: I believe this was so that the process could be initiated manually for testing and shouldn't be in the final version of the file. @jayvaznewm is that correct?

dmkirshon (approved):

.github/workflows/aws-hosting-deploy-published-release-newm-studio.yml
escobarjonatan: James will comment on this shortly but confirmed through chat we are good to merge

PR #672, opened 2024-06-02
Renders marketplace artist banner

scandycuz:


apps/marketplace/src/components/ArtistSpotlight.tsx
```diff
@@ -4,7 +4,6 @@ import Artists from "./Artists";
 import { useGetArtistsQuery } from "../modules/artist";
 
 const ArtistSpotlight: FunctionComponent = () => {
-  // TODO: update params to only return artists with sales when API updated
   const { isLoading, data: artists = [] } = useGetArtistsQuery({
```
scandycuz, line 7: The "Get Marketplace Artists" endpoint now only returns artists who have created at least one sale.

packages/types/src/lib/song.ts
```diff
@@ -56,7 +56,7 @@ export interface Song {
   readonly parentalAdvisory?: string;
   readonly phonographicCopyrightOwner: string;
   readonly phonographicCopyrightYear: string;
-  readonly price?: string; //TODO: Update name after key added to song endpoint
+  readonly price?: string; // TODO: Update name after key added to song endpoint
```
escobarjonatan: The reason why i added this comment to the interface was because i was just guessing what the key for the price would be and looks like i had a lucky guess that ended up matching what BE did. This comment can be deleted

escobarjonatan (approved):

packages/types/src/lib/song.ts
scandycuz: Removed it.

PR #675, opened 2024-06-03
FEATURE: Marketplace Tab V1 Implementation

escobarjonatan:
This PR will be an aggregate of all things related to this Epic: [Marketplace Tab V1 Implementation](https://projectnewm.atlassian.net/browse/STUD-237)

- [x] [STUD-247: Create Marketplace tab outline and make available only for tracks in "Distributed and Minted" or "Minted and Released" status.](https://projectnewm.atlassian.net/browse/STUD-247)
- [x] [STUD-248: Implement logic checkpoint to determine whether there is already an active sale (or one scheduled to be created) on the Marketplace for the track..](https://projectnewm.atlassian.net/browse/STUD-248)
- [x] [STUD-249: Implement UI for when there is already an active stream token sale on the Marketplace.](https://projectnewm.atlassian.net/browse/STUD-249)
- [x] [STUD-251: A wallet containing the track's stream tokens must be connected in order to create a new sale.](https://projectnewm.atlassian.net/browse/STUD-251)
- [x] [STUD-252: Implement UI to create a stream token sale.](https://projectnewm.atlassian.net/browse/STUD-252)

This will be out of scope:
- [ ] [STUD-250: Implement UI for when there is a stream token sale scheduled to be created on the Marketplace.](https://projectnewm.atlassian.net/browse/STUD-250)


scandycuz (approved):
Looks great!

apps/studio/src/modules/sale/thunks.ts
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * Starts a sale for a song.
+ */
+export const startSale = createAsyncThunk(
+  "sale/start",
+  async (body: StartSaleThunkRequest, { dispatch }) => {
+    try {
+      const wallet = await enableWallet();
+      const changeAddress = encodeAddress(await wallet.getChangeAddress());
```
scandycuz: I'd recommend using the `@newm.io/cardano-dapp-wallet-connector` `getWalletChangeAddress` util function here instead. It handles encoding the address, and I believe some wallets can return an array of instead of a single string, which the util also handles.

PR #679, opened 2024-06-08
STUD-251: Add wallet connect to marketplace tab

escobarjonatan:
No Connected wallet:
image

Connected wallet. NO stream tokens:
image

TEMP: Has stream tokens:
image


apps/studio/src/pages/home/library/Marketplace/Marketplace.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { MarketplaceTabSkeleton } from "../../../../components";
+import { SongRouteParams } from "../types";
+import { useGetUserWalletSongsThunk } from "../../../../modules/song";
+
+export const Marketplace = () => {
+  const { songId } = useParams<"songId">() as SongRouteParams;
+  const [getUserWalletSongs, { data: walletSongsResponse, isLoading }] =
+    useGetUserWalletSongsThunk();
+
+  const hasStreamTokens =
```
escobarjonatan: Technically if the response has songs length then it should be equal to what was requested but added an exact match just in case? Not sure if anyone has a better way to check for this 👀

apps/studio/src/modules/song/types.ts
```diff
@@ -1,3 +1,4 @@
+import { off } from "process";
```
scandycuz: Was this an accidental import?

scandycuz (commented):
LGTM, just one minor comment

apps/studio/src/pages/home/library/Marketplace/Marketplace.tsx
```diff
@@ -0,0 +1,50 @@
+import { useEffect } from "react";
+import { useParams } from "react-router";
+import { Box, Typography } from "@mui/material";
+import theme from "@newm-web/theme";
+import { ConnectWallet } from "./ConnectWallet";
+import { MarketplaceTabSkeleton } from "../../../../components";
+import { SongRouteParams } from "../types";
+import { useGetUserWalletSongsThunk } from "../../../../modules/song";
+
+export const Marketplace = () => {
```
dmkirshon: Not a big deal, but with the monorepo and our references growing, can this be renamed to `MarketplaceTab`, and any other related refactors? Just noticed the `MarketplaceTabSkeleton` so makes sense with consistency as well.

apps/studio/src/pages/home/library/Marketplace/ConnectWallet.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <Stack>
+      <Alert>
+        <Typography
+          color={ theme.colors.blue }
+          fontWeight={ 500 }
+          variant="subtitle1"
+        >
+          { wallet
```
dmkirshon: Once the wallet was connected I had to either refresh the page or switch back and forth between tabs to refresh the component, otherwise it just shows that my wallet doesn't have stream tokens which wasn't valid. Maybe pass a state down from `MarketplaceTab` to allow for a rerender of the Tab?

dmkirshon (changes requested):
Just two minor things: a nomenclature refactor request and tab rerender issue

apps/studio/src/pages/home/library/Marketplace/ConnectWallet.tsx
escobarjonatan: I was planning to take care of this in the next iteration, i solved it now tho. It just needed to do the check again if the wallet had changed.

dmkirshon (approved):
LGTM!

PR #681, opened 2024-06-11
MRKT-27: Update for song previews to prevent concurrent playback  

dmkirshon:
Add:
- Modules package directory for Redux state management integration

Update:
- Integrate state management into `audio.tsx` to prevent two different components playing song previews at the same time through the hook


packages/utils/src/lib/audio.ts
```diff
@@ -1,22 +1,31 @@
...
  * Hook used to play an audio file using the Howl library.
  */
 export const usePlayAudioUrl = () => {
-  const [audio, setAudio] = useState<Howl>();
-  const [audioUrl, setAudioUrl] = useState<string>();
-  const [isAudioPlaying, setIsAudioPlaying] = useState<boolean>(false);
-  const [audioProgress, setAudioProgress] = useState<number>(0);
+  const dispatch = useDispatch();
+  const { audio, audioUrl, audioProgress, isAudioPlaying }: AudioState =
+    useSelector((state: { audio: AudioState }): AudioState => state.audio);
```
dmkirshon: In our main apps we use a custom `useAppSelector` hook but without `store.ts` and access to the `RootState` I wasn't able to integrate that here. Is there a better way to do this? Move over `store.ts` and other redux tooling to the packages directory?

packages/modules/src/lib/marketplace/audio/slice.ts
```diff
@@ -0,0 +1,53 @@
+import { PayloadAction, createSlice } from "@reduxjs/toolkit";
+import { AudioState } from "./types";
+
+const initialState: AudioState = {
+  audio: undefined,
```
dmkirshon: Is it better to set this as null?

packages/utils/src/lib/audio.ts
```diff
@@ -63,31 +74,29 @@ export const usePlayAudioUrl = () => {
       // if not currently selected song, stop playing
       if (audio?.playing()) {
         audio.stop();
+        dispatch(resetAudioState());
```
dmkirshon, line 76: I was having some issues with order of operations for the Howl `onstop` callback when switching between song previews, but will revisit one more time to see if I can refactor the state resets.

scandycuz (approved):

PR #683, opened 2024-06-13
Move Audio Hook to new audio package

dmkirshon:
Moves `usePlayAudioUrl` hook to own package. Refactors state management to utilize `useContext` and `useReducer`. Overall done to simplify and make hook more modular and self-contained.


packages/audio/src/lib/actions.ts
```diff
@@ -0,0 +1,44 @@
+import {
+  AudioState,
+  ResetAudioStateAction,
+  SetAudioAction,
+  SetAudioProgressAction,
+  SetAudioUrlAction,
+  SetIsAudioPlayingAction,
+} from "./types";
```
scandycuz, line 9: This works, however, another way to create and define the actions in less code is to use `createAction` from `redux/toolkit`. E.g.

```
import { createAction } from "@reduxjs/toolkit";

...

const SET_AUDIO = "setAudio"

const setAudio = createAction<Howl | undefined>(SET_AUDIO)
```

instead of 

```
export interface SetAudioAction {
  readonly payload: Howl | undefined;
  readonly type: typeof SET_AUDIO;
}

export const setAudio = (payload: AudioState["audio"]): SetAudioAction => {
  return {
    payload,
    type: "setAudio",
  };
};
```

scandycuz (approved):
Had a minor comment but LGTM!

packages/audio/src/lib/reducer.ts
```diff
@@ -0,0 +1,27 @@
...
+  audioUrl: "",
+  isAudioPlaying: false,
+};
+
+const reducer = (state: AudioState, action: Action) => {
+  switch (action.type) {
+    case "resetAudioState":
+      return initialState;
+    case "setAudio":
+      return { ...state, audio: action.payload };
```
scandycuz: One thing that can be good to do, just to ensure all of the string values are accurate, is to use the action type variables here. E.g.

```
import { SET_AUDIO } from "./types";

...

case SET_AUDIO:
    return { ...state, audio: action.payload };
```

PR #684, opened 2024-06-20
MRKT-32: purchase stream tokens

scandycuz:


apps/marketplace/src/modules/sale/thunks.ts
```diff
@@ -0,0 +1,72 @@
...
+
+      const { orderId, amountCborHex } = orderResp.data;
+      const wallet = await enableWallet();
+      const changeAddress = await getWalletChangeAddress(wallet);
+      const utxoCborHexList = await wallet.getUtxos(amountCborHex);
+
+      if (!utxoCborHexList) {
+        throw new Error(
+          `Insufficient NEWM tokens in wallet. Please add NEWM tokens to
+          your wallet and try again.`
```
escobarjonatan, line 30: Minor, but this can be a regular string and not a template string 🤷🏽‍♂️

escobarjonatan: Also, the message here makes sense, but how does a user add NEWM tokens to their wallets? Should we have a section explaining the options to buy NEWM or is this a known thing?

escobarjonatan (approved):

apps/marketplace/src/modules/sale/thunks.ts
scandycuz: It's just preference, but the template string is to avoid concatinating strings on two lines, e.g.
```
"Insufficient NEWM tokens in wallet. Please add NEWM tokens to " +
"your wallet and try again."
```

scandycuz: I believe the plan is to provide a widget that will let users easily purchase NEWM tokens, but there aren't mock-ups for it currently.

PR #689, opened 2024-06-25
STUD-78-Consolidate-Distribute-Minting-form-logic-for-uploaded-songs

dmkirshon:
Refactor:
- Remove alerts that aren't visible and don't apply to in-process/completed mint
- Remove flow for contract signing and minting initialization as view-song flow doesn't serve this purpose
- miscellanious clean up on component nomenclature and MUI usage

Fix:
- Owner percentage yup validation issue due to string type

### DIFF
--- a/apps/studio/src/common/formUtils.ts
+++ b/apps/studio/src/common/formUtils.ts
@@ -211,7 +211,7 @@ export const commonYupValidation = {
           if (!owners) return false;
 
           const percentageSum = owners.reduce((sum, owner) => {
-            return sum + owner.percentage;
+            return sum + Number(owner.percentage);
           }, 0);
 
           return percentageSum === 100;
--- a/apps/studio/src/pages/home/library/MintSong.tsx
+++ b/apps/studio/src/pages/home/library/MintSong.tsx
@@ -1,11 +1,16 @@
-import { AlertTitle, Box, Button as MUIButton, Stack } from "@mui/material";
+import {
+  AlertTitle,
+  Box,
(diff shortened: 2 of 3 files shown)


apps/studio/src/common/formUtils.ts
```diff
@@ -211,7 +211,7 @@ export const commonYupValidation = {
           if (!owners) return false;
 
           const percentageSum = owners.reduce((sum, owner) => {
-            return sum + owner.percentage;
+            return sum + Number(owner.percentage);
```
dmkirshon, line 214: the `TextInputField` for owner is of type number but the value of the text input is still a string. The calculation done here was adding two strings together and causing the percentage sums to error.

apps/studio/src/pages/home/library/MintSong.tsx
```diff
@@ -302,206 +212,81 @@ const MintSong = () => {
...
+                    windowWidth && windowWidth > theme.breakpoints.values.md
+                      ? "compact"
+                      : "default"
+                  }
+                  onClick={ () => navigate(-1) }
+                >
+                  Cancel
+                </Button>
+
+                { dirty && (
```
dmkirshon, line 275: This conditional was previously to toggle the button visibility for an update collab details or move onto the next step for minting. As minting flow is removed through this PR, does it still make sense to have this button appear only when the collab details are modified?

escobarjonatan (approved):

scandycuz (approved):

PR #694, opened 2024-07-03
MRKT-93: format stream token NEWM values

scandycuz:
- Converts sale NEWM cost to correct value
- Updates NEWM conversion util to display two decimals places instead of one to reflect required precision for individual stream token prices
- Updates song card approximate USD value from two to three decimal places to match Studio [Create sale mock-ups](https://www.figma.com/design/zrn01ZR2zFeLiKBebEYI1v/NEWM-Studio---Final-Designs?node-id=10091-734&t=5KOyOWRyvxFYG3jL-0) and required precision for individual stream token prices

Screen Shot 2024-07-02 at 7 36 07 PM

### DIFF
--- a/packages/utils/src/lib/crypto.ts
+++ b/packages/utils/src/lib/crypto.ts
@@ -1,5 +1,3 @@
-import currency from "currency.js";
-
 /**
  * Formats a numerical NEWM amount with the correct decimal
  * places and symbol.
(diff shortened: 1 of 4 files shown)


packages/utils/src/lib/crypto.ts
```diff
@@ -9,7 +9,18 @@ export const formatNewmAmount = (amount?: number, includeSymbol = true) => {
...
+
+/**
+ * Formats a numerical USD amount to three decimal places
+ * rather than the standard two, based on the exchange rate
+ * for NEWM to USD.
+ */
+export const formatUsdAmount = (amount?: number) => {
+  if (!amount) return "";
+
+  return currency(amount, { precision: 3 }).format();
```
escobarjonatan: Should the precision be a var and set to 3 as default?

escobarjonatan (approved):

packages/utils/src/lib/crypto.ts
scandycuz: Probably a good idea

PR #696, opened 2024-07-08
Stud 249 UI for active sale

escobarjonatan:
This includes:
- Create sale UI (minus summary modal)
- Create/End sale endpoint additions
- Active sale UI with partial endpoint logic to cancel sale (currently manual process)
- Pinging for sale endpoint on sale end/start 

https://github.com/user-attachments/assets/08fb9f37-3805-4709-bb82-063d6d6cd936


apps/studio/src/api/utils.ts
```diff
@@ -105,18 +105,12 @@ export const getAuthHeaders = (api: BaseQueryApi) => {
...
   const action = recaptchaEndpointActionMap[endpoint] || endpoint;
 
-  if (!isLoggedIn) {
-    return {
-      "g-recaptcha-platform": "Web",
-      "g-recaptcha-token": await executeRecaptcha(action),
-    };
-  }
-
-  return {};
```
escobarjonatan, line 119: @scandycuz Having this prevented us from calling the get sales marketplace endpoint in studio. Let me know your thoughts on removing it

apps/studio/src/App.tsx
```diff
@@ -50,6 +52,8 @@ const App = () => {
             <ProgressBarModal />
             <UpdateWalletAddressModal />
             <WalletEnvMismatchModal />
+            <PingSaleStart />
+            <PingSaleEnd />
```
escobarjonatan, line 56: These components will handle polling for new sales when one gets created/removed in order to display an accurate UI state of the sale

apps/studio/src/api/utils.ts
scandycuz: Should be fine to send the headers for all requests

scandycuz (approved):
LGTM!

PR #700, opened 2024-07-23
MRKT-94: adds unsupported browser banner to apps

scandycuz:
Adds "Unsupported browser banner" to Studio, Marketplace, and Mobile Wallet Connector apps

Screen Shot 2024-07-23 at 12 06 01 AM

### DIFF
--- a/packages/components/src/lib/UnsupportedBrowserBanner.tsx
+++ b/packages/components/src/lib/UnsupportedBrowserBanner.tsx
@@ -37,10 +37,12 @@ const UnsupportedBrowserBanner: FunctionComponent = () => {
     <Stack
       alignItems="center"
       justifyContent="center"
+      position="relative"
       py={ 2 }
       sx={ {
         backgroundColor: theme.colors.grey600,
       } }
+      zIndex={ 1000000 }
     >
       <Container maxWidth="lg">
         <Alert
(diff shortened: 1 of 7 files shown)


packages/components/src/lib/UnsupportedBrowserBanner.tsx
```diff
@@ -0,0 +1,71 @@
...
+    <Stack
+      alignItems="center"
+      justifyContent="center"
+      py={ 2 }
+      sx={ {
+        backgroundColor: theme.colors.grey600,
+      } }
+    >
+      <Container maxWidth="lg">
+        <Alert
```
dmkirshon, line 48: Looks like the `z-index` needs to be updated as it hides behind the sidebar.

In Firefox:
image

Also seen by changing user agent in Chrome to Opera browser:
image

dmkirshon (approved):
LGTM, just one issue with the z-index in Studio

packages/components/src/lib/UnsupportedBrowserBanner.tsx
```diff
@@ -0,0 +1,73 @@
+import { Container, Stack, useTheme } from "@mui/material";
+import { Alert, Button, Typography } from "@newm-web/elements";
```
escobarjonatan, line 2: Typography should come from MUI

packages/components/src/lib/UnsupportedBrowserBanner.tsx
scandycuz: Good catch, made the update.

PR #703, opened 2024-07-29
STUD-287: Update limit on song description characters

escobarjonatan:


apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
```diff
@@ -379,6 +383,13 @@ const BasicSongDetails: FunctionComponent<BasicDonDetailsProps> = ({
             placeholder="Tell us about your song"
             ref={ descriptionRef }
           />
+          <Typography
+            component="span"
+            style={ { marginTop: "4px", textAlign: "end" } }
+            variant="subtitle2"
+          >
+            { `${values.description?.length}/${SONG_DESCRIPTION_MAX_CHARACTER_COUNT}` }
+          </Typography>
```
dmkirshon: The error message appearing and nudging the character count is a little jarring
image

Can we move it within the `TextAreaField.tsx` to `TextArea.tsx` to be in line with the error message?
https://github.com/projectNEWM/newm-web/blob/5d6cab6ab395621537b8facdcd0ee4b73ac9cbf2/packages/elements/src/lib/TextArea.tsx#L168-L169

dmkirshon (changes requested):
LGTM, just requesting a change in the character count placement to by in sync with error message.

apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
scandycuz: Good catch, it would be ideal to have these on the same line.

escobarjonatan: awk, tested both but not together or didn't notice it. Fixed it now:

image
image

apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
```diff
@@ -374,6 +378,8 @@ const BasicSongDetails: FunctionComponent<BasicDonDetailsProps> = ({
           </Stack>
 
           <TextAreaField
+            characterCountLimit={ SONG_DESCRIPTION_MAX_CHARACTER_COUNT }
+            currentCharacterCount={ values.description?.length }
```
dmkirshon, line 382: Not a show stopper, but couldn't this be obtained within the `TextArea.tsx` through the `value` prop? Is there a reason to prop drill it?

dmkirshon (approved):
LGTM, just a question about the character count prop.

apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
escobarjonatan: I thought about it but think this is a cleaner solution since it can give the flexibility to omit characters if needed? for example, you could use a text area with validation to only count words over text and this gives you the ability to add your own custom logic. You also have controlled vs uncontrolled inputs and think destructuring the value and adding it would interfere with uncontrolled

PR #704, opened 2024-07-30
STUD-292: Replace NEWM Typography with MUI for consistency

escobarjonatan:
Did a smoke test with it and seemed all good, previous component was wrapping typography and passing in a color from theme.

image
image
image
image
image


dmkirshon (approved):
LGTM

PR #705, opened 2024-07-31
STUD-54: Update socials validation

escobarjonatan:


apps/studio/src/common/regex.ts
```diff
@@ -35,7 +35,7 @@ const generateRegexForDomainsWithPath = (
     .map((domain) => domain.replace(/\./g, "\\."))
     .join("|");
   return new RegExp(
-    `^(https?:\\/\\/(www\\.)?)?(${domainsPattern})(\\/[^\\s]+)`,
+    `^(https?:\\/\\/)?(www\\.)?(${domainsPattern})(\\/[^\\s]+)`,
```
escobarjonatan, line 38: Request:

>the artist can enter [https://www.instagram.com/username](https://instagram.com/username) OR [instagram.com/username](http://instagram.com/username) BUT NOT [www.instagram.com/username.](http://www.instagram.com/username)

>Can we make it so that if the user enters a [www.instagram.com/username](http://www.instagram.com/username) value the app accepts it?

To ensure `www.instagram.com/username` is valid, modify the optional so that it correctly handles the presence of www. for URLs.

dmkirshon (approved):
LGTM

scandycuz (approved):

PR #706, opened 2024-07-31
STUD-302-Integrate-feature-flag-into-wallet-enhancement-feature-branch

dmkirshon:
Add:
- Feature Flag for wallet royalties claiming and enhancements
- Logic flow and UI for current Studio state coming soon feature

### DIFF
--- a/apps/studio/src/pages/home/wallet/Wallet.tsx
+++ b/apps/studio/src/pages/home/wallet/Wallet.tsx
@@ -16,12 +16,17 @@ import {
 } from "@mui/material";
 import theme from "@newm-web/theme";
 import { useConnectWallet } from "@newm.io/cardano-dapp-wallet-connector";
+import { Button } from "@newm-web/elements";
 import { UnclaimedRoyalties } from "./UnclaimedRoyalties";
 import Portfolio from "./Portfolio";
 import Transactions from "./Transactions";
 import { NoPendingEarnings } from "./NoPendingEarnings";
 import { NoConnectedWallet } from "./NoConnectedWallet";
-import { resetWalletPortfolioTableFilter } from "../../../modules/ui";
+import { useGetStudioClientConfigQuery } from "../../../modules/content";
+import {
+  resetWalletPortfolioTableFilter,
+  setIsConnectWalletModalOpen,
+} from "../../../modules/ui";
 import { useAppDispatch } from "../../../common";
 import { DisconnectWalletButton } from "../../../components";
 import { useGetUserWalletSongsThunk } from "../../../modules/song";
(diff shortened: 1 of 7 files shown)


apps/studio/src/pages/home/wallet/Wallet.tsx
```diff
@@ -89,76 +99,120 @@ const Wallet: FunctionComponent = () => {
...
-            display: "flex",
-            justifyContent: "space-between",
-            mb: 5,
-          } }
-        >
-          <Typography fontWeight={ 800 } variant="h3">
-            WALLET
-          </Typography>
-          <DisconnectWalletButton />
+  if (isClientConfigLoading) return;
```
dmkirshon, line 104: I made this a blank return for the wallet page during loading of the feature flag. I was thinking of putting the wallet title and or connect button but the different views of this page, between the feature and current state in Studio, are different enough that seeing copy/elements doesn't make sense right now. Have no issue changing the default.

apps/studio/src/pages/home/wallet/Wallet.tsx
```diff
@@ -89,76 +99,120 @@ const Wallet: FunctionComponent = () => {
...
-              } }
-              value={ tab }
-              onChange={ handleChange }
-            >
-              <Tab aria-controls="tabpanel-0" id="tab-0" label="PORTFOLIO" />
-              <Tab aria-controls="tabpanel-1" id="tab-1" label="TRANSACTIONS" />
-            </Tabs>
+      </Container>
+    );
+  } else {
```
dmkirshon, line 145: This is where the new Wallet Enhancement features come into play, this is the same code that was already part of this enhancement.

apps/studio/src/pages/home/wallet/Wallet.tsx
```diff
@@ -89,76 +99,120 @@ const Wallet: FunctionComponent = () => {
...
+                Connect Wallet
+              </Button>
+            ) }
+          </Box>
+          <Box mt={ 5 } pb={ 5 }>
+            <Typography variant="body1">
+              Wallet Portfolio and Transactions Coming Soon
+            </Typography>
+          </Box>
         </Box>
```
dmkirshon: Added this section to handle when the feature flag is false so it still covers the connect wallet purpose for the Wallet page in Studio.

apps/studio/src/pages/home/wallet/Wallet.tsx
escobarjonatan: I wonder if we can just create a skeleton and use this when something is loading in this page? Just a thought. I'm ok with going with what you have

escobarjonatan (approved):

apps/studio/src/pages/home/wallet/legacyWallet/LegacyTableDropdownSelect.tsx
dmkirshon, line 1: I imported this component as I believe it was the only one changed for the new feature.

apps/studio/src/pages/home/wallet/Wallet.tsx
dmkirshon: The current skeletons are buried in the lower components, mainly for the table loading, I feel like it would be a bit of a mess to try and pull them out and separate them per feature flag logic. Up for suggestions though, but like I mentioned the two states are quite different so they would each require their own skeletons.

scandycuz (approved):
LGTM

PR #710, opened 2024-08-21
MRKT-135: references NEWM token testnet details in staging

scandycuz:


escobarjonatan (approved):
Much better shared lol

scandycuz:
We should likely make a new package, specific for constants.

PR #716, opened 2024-08-28
[STUD-321, MRKT-129]: updates sales for newm and usd currencies

scandycuz:
Updates Studio to allow for creating sales in different currencies.


scandycuz:
Marking as a draft, as the back-end will be updated to return the NEWM token price for all sales, which the front-end can reference.

scandycuz:
Closed in favor of https://github.com/projectNEWM/newm-web/pull/717

PR #717, opened 2024-08-29
[STUD-321, MRKT-129, MRKT-139] Finalize initial marketplace sales

scandycuz:
- Updates Studio to allow for creating sales in different currencies.
- Updates Marketplace to allow for sales created with USD and NEWM currency
- Implements confirmation modal for stream token purchases

### DIFF
--- a/apps/marketplace/src/modules/sale/thunks.ts
+++ b/apps/marketplace/src/modules/sale/thunks.ts
@@ -51,8 +51,7 @@ export const purchaseStreamTokens = createAsyncThunk(
 
       dispatch(
         setToastMessage({
-          message: `Stream token purchase was successful, please allow up 
-            to 5 minutes for the transaction to complete.`,
+          message: "Stream token purchase was successful.",
           severity: "success",
         })
       );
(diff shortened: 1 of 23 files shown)


scandycuz:
Marked as a draft, as the back-end will be updated to return the NEWM token price for all sales, which the front-end can reference.

apps/marketplace/src/modules/sale/thunks.ts
```diff
@@ -51,8 +51,7 @@ export const purchaseStreamTokens = createAsyncThunk(
 
       dispatch(
         setToastMessage({
-          message: `Stream token purchase was successful, please allow up 
-            to 5 minutes for the transaction to complete.`,
+          message: "Stream token purchase was successful.",
```
escobarjonatan, line 54: Why remove the 5 minutes? lol I know it's fast but not instant

escobarjonatan (approved):

apps/marketplace/src/modules/sale/thunks.ts
scandycuz: It's because the new purchase confirmation modal has a disclaimer about it taking a few minutes, and I figured it probably wasn't necessary to repeat it. I reached out to product about clarification around the language though, and we can always see what users' feedback is during internal testing.

PR #734, opened 2024-09-24
[MRKT-155, STUD-339]: clears wallet on disconnect for all apps

scandycuz:
Ensures data displayed in wallet UI is cleared when wallet is disconnected.


apps/studio/src/components/ConnectWalletModal.tsx
```diff
@@ -40,11 +56,47 @@ const ConnectWalletModal: FunctionComponent = () => {
...
+   * Gets an address from the wallet and updates the Redux state.
+   */
+  useEffect(() => {
+    if (wallet) {
+      getAddress((value) => {
+        dispatch(setWalletAddress(value));
+      });
+    }
+  }, [wallet, getAddress, dispatch]);
+
```
scandycuz, line 93: Moved the logic to set wallet balances for Studio here, just so all of the wallet balance logic is in the same place.

escobarjonatan (approved):

PR #738, opened 2024-10-01
STUD-331: Add earnings claim modal and required endpoints

escobarjonatan:
https://github.com/user-attachments/assets/828738f7-7cd4-48ea-bde0-ef8ba0266613


apps/studio/src/modules/wallet/types.ts
```diff
@@ -3,3 +3,58 @@ export interface WalletState {
...
+  readonly amountCborHex: string;
+  readonly earnings: Earning[];
+  readonly totalClaimed: number;
+};
+
+export type PostEarningsRequest = {
+  readonly changeAddress: string;
+  readonly utxoCborHexList: string[];
+  readonly walletAddress: string;
+};
```
scandycuz: Was defining these as types instead of interfaces intentional?

escobarjonatan: negative, I switch between types and interfaces in different projects. made them interface for consistency here

scandycuz: I was thinking the `GetEarningsResponse` and `PostEarningsRequest` types would be more consistent to have as interfaces.

scandycuz (approved):
Had a minor comment, but LGTM!

PR #757, opened 2024-10-22
MRKT-154: incorrect NEWM purchase fee

scandycuz:
- Corrects NEWM fee amount for purchase stream tokens modal.

### Also

- Updates decimal place for individual stream token USD price from 3 to 4 places, as 3 was determined to be insufficient for smaller sales. A longer term fix will be to dynamically determine the decimal places based on the amount.

Screen Shot 2024-10-21 at 11 46 23 PM

### DIFF
--- a/apps/marketplace/src/components/modals/PurchaseStreamTokensModal.tsx
+++ b/apps/marketplace/src/components/modals/PurchaseStreamTokensModal.tsx
@@ -7,7 +7,10 @@ import {
   formatNewmAmount,
   formatUsdAmount,
 } from "@newm-web/utils";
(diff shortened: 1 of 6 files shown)


apps/marketplace/src/components/modals/PurchaseStreamTokensModal.tsx
dmkirshon, line 1: I'm still seeing some situations where there are $0.00 values. Is this already something in discussion with product? These also might just be edge cases of small value per token and small bundles.

image

---

image

dmkirshon (approved):
LGTM, just had a question for potential edge cases dealing with the decimal placements and $0.00 values.

PR #758, opened 2024-10-23
MRKT-161: ensures numeric input values stored in formik as numbers

scandycuz:


packages/elements/src/lib/form/TextInputField.tsx
```diff
@@ -8,14 +8,36 @@ const TextInputField: ForwardRefRenderFunction<
...
+          form.setFieldValue(field.name, formattedValue);
+        };
+
+        return (
+          <TextInput
+            errorMessage={ meta.touched ? meta.error : "" }
+            ref={ ref }
+            { ...field }
+            { ...props }
+            onChange={ handleChange }
```
escobarjonatan, line 37: If an onchange is passed I believe this will override that. Do we have instances with an onchange/handle change we need to account for?

escobarjonatan (approved):

packages/elements/src/lib/form/TextInputField.tsx
scandycuz: There shouldn't be any use case where `onChange` would be passed to a Formik field. For inputs where an `onChange` prop is passed, the non-Formik component can be used.

PR #765, opened 2024-10-28
STUD-146: Update transactions tab to include wallet and user transactions

escobarjonatan:
https://github.com/user-attachments/assets/19dbe616-6a74-483b-bf11-648e6f18f543


apps/studio/src/pages/home/wallet/transactions/Transaction.tsx
dmkirshon, line 1: Can this one be renamed to `SingleTransaction` or something similar to get a little difference from the `Transactions`. `Transactions` can be changed instead or as well. Just very similar in name.

packages/utils/src/lib/crypto.ts
```diff
@@ -89,6 +89,10 @@ export const convertNewmiesToNewm = (newmies: number): number => {
   return newmies / LOVELACE_CONVERSION;
 };
 
+export const convertAdaiesToAda = (adaies: number): number => {
+  return adaies / LOVELACE_CONVERSION;
+};
+
```
dmkirshon: I think at this level it is a `convertLovelaceToAda`, is it not? I see the pattern though :+1:

apps/studio/src/pages/home/wallet/transactions/Transactions.tsx
dmkirshon, line 1: Just to see if I understand the business logic. The wallet transactions initially displayed will be all the earnings that have been claimed via the connected wallet + First 25 Minted song. Each scroll to the end of the transaction display list will fetch the next 25 descending sorted Minted songs and combine it into the sorted claimed+minted date list. Is this correct?

Don't need to pursue this Just a bit of an edge case and UX design, I'm sure this will be fine for now with the minimal minting and claiming. What happens when there are previous earning claims and then more than 25 new songs minted since those earnings have been claimed. Would some of the next offset fetched minted songs interlace into the already displayed claimed earnings dates?

dmkirshon (commented):
Still want to see if I can load some of the transactions through my account, currently seeing none. Otherwise changes look good, few comments.

apps/studio/src/pages/home/wallet/transactions/Transaction.tsx
escobarjonatan: Sure thing, fixed in the next push

packages/utils/src/lib/crypto.ts
escobarjonatan: yes, however newm/newmies & lovelace/ada. i wanted to keep it in that pattern. I think I'm going to stick to that unless you feel strongly about the lovelace

apps/studio/src/pages/home/wallet/transactions/Transactions.tsx
escobarjonatan: The logic is correct as you have described it.
In the edge case you have described if I understood it correctly it will:
First, fetch all of the claimed earnings, say 20. It will also fetch the first 25 artist songs... wherever the first 25 songs land in date order combined with the claims is where they will display. If first 25 are all newer than the earnings, it will display those and as the user scrolls down it will arrange them where they belong. I think this will be ok for now until we revisit the pagination for our earning claims :)

packages/utils/src/lib/crypto.ts
scandycuz: I'd have to agree that we should refer to the smaller ADA denomination as "lovelaces", since "lovelaces" are the official term.

scandycuz (approved):
LGTM

PR #776, opened 2024-11-02
STUD-361: Update wallet portfolio to new requirements

escobarjonatan:
https://github.com/user-attachments/assets/ae909cb1-818b-4f9d-a67b-2574753674f9


apps/studio/src/modules/song/api.ts
```diff
@@ -715,6 +715,7 @@ export const {
   useGetSongQuery,
   useGetSongStreamQuery,
   useGetSongsQuery,
+  useLazyGetSongsQuery,
```
escobarjonatan, line 718: @scandycuz @dmkirshon, Sharing this for awareness:

In RTK Query, `useQuery` and `useLazyQuery` are hooks for fetching data, each suited to different scenarios:

`useQuery`: **Automatically** initiates a data fetch when the component mounts, making it ideal for scenarios where data is needed immediately upon rendering. 

`useLazyQuery`: Provides a trigger function that allows you to **manually** start the data fetch at a later time, offering greater control over when the fetch occurs. 

Their doc has some other queries that I think could be helpful in their own scenario. [RTK Query documentation on queries](https://redux-toolkit.js.org/rtk-query/usage/queries).

escobarjonatan: In this case i needed to trigger a fetch for new data on demand for when the user modified the filters

scandycuz: If you still want fetch data automatically in addition to programatically, another option is to use the `refetch` function returned from the `useQuery` hook. E.g. `const { data, refetch } = useExampleQuery()`.

scandycuz (approved):

apps/studio/src/modules/song/api.ts
escobarjonatan: @scandycuz I tried that and I would get mixed results where songs would display twice so somehow previous results were not clearing

PR #790, opened 2024-11-09
STUD-351: Add Launch Darkly and apply a maintenance mode to be activated on demand

escobarjonatan:
https://github.com/user-attachments/assets/54d59802-6702-4d68-b8ee-6a082509dd98


apps/studio/src/components/LDUserUpdater.tsx
```diff
@@ -0,0 +1,42 @@
+import { useEffect } from "react";
```
escobarjonatan, line 1: This component is what adds the user information to LD once they log in, we can then use that info to target certain features by user

docs/launch-darkly-feature-flags.md
```diff
@@ -0,0 +1,40 @@
+# Feature Flag Naming Conventions for LaunchDarkly
```
escobarjonatan, line 1: Decided to start off this doc to try to follow a standard since we will be sharing the space with the rest of the NEWM platforms, feedback is welcomed as always.

global.css
```diff
@@ -1,56 +1,60 @@
 html {
```
escobarjonatan, line 1: Not sure why git thinks this is a repath but this is the monorepo global.css to be used in every new app that gets created. you might be wondering, how did I come up with this.... GREAT question. I saw what we had in all of the apps and combined into 1, only what made sense. I moved some styles to app-specific spaces. I created a [ticket](https://projectnewm.atlassian.net/browse/STUD-372) to create a new package if needed. This was the easiest path I found.

docs/launch-darkly-feature-flags.md
scandycuz: Looks good to me

scandycuz (approved):

dmkirshon (approved):

cristhianescobar (approved):
LGTM!

PR #808, opened 2024-11-14
chore: Update destroy workflows to checkout internal dependencies to destroy the url environment

escobarjonatan:


.github/workflows/cdk/package-lock.json
dmkirshon, line 1: package-lock file changing again, specific reason why it updated outside of an npm package update?

dmkirshon (approved):
LGTM, just curious about the package-lock update

.github/workflows/cdk/package-lock.json
escobarjonatan: I believe that was part of the problem. It didn't install dependencies because the lock and the install didn't match. I did an install locally and it modified the lock so this is kinda the "fresh start" iteration. Going forward it should not change unless a dependency changes. Also, I made it so that all workflows in the cdk share the same dependencies and keep them consistent.

PR #813, opened 2024-11-15
MRKT -189, STUD-355, STUD-356 updates to improve currency UI

dmkirshon:
Updated:
- Increase font weight of input field titles and subtexts
- Increase default decimal precision for NEWM currency value
- Moved ≈ estimate symbol from jsx locations to currency formatter as an option, allows for "less than ..." currency statement return without estimation symbol

Add:
- Add dynamic decimal precision on studio marketplace tab for percentage and currency values
- Add a return string for "less than .." of the default precision value for NEWM and USD currency formatting, present in studio and marketplace modals and song cards (not used if precision is modified to lower than default)


packages/components/src/lib/earnings/EarningsSummaryModal.tsx
```diff
@@ -77,8 +77,11 @@ const EarningsSummaryModal: FunctionComponent<
                 <Typography variant="subtitle1">Earnings</Typography>
                 <Typography variant="h4">
                   <Typography component="span" mr={ 0.5 } variant="subtitle2">
-                    (≈{ " " }
-                    { formatUsdAmount(unclaimedEarningsInUSD, { precision: 2 }) }){ " " }
+                    (
+                    { formatUsdAmount(unclaimedEarningsInUSD, {
+                      includeEstimateSymbol: true,
+                    }) }
+                    ){ " " }
```
dmkirshon, line 84: Do we know why these extra spaces are located everywhere with the currency format jsx? Existed before I added this in but I think one or two were added in for consistency I believe. Do we have issue removing them or are they in place for text formatting?

packages/utils/src/lib/crypto.ts
dmkirshon, line 1: There is quite a bit of redundancy across these currency formatters. Are we ok to refactor this on another ticket and have the currency as an option for a generic formatter? Or is it better to keep these separated?

I also noted we have some `formatPriceToDecimal` functions from the pricing plan/summary that seem to serve a similar purpose that could also be merged in.

packages/utils/src/lib/crypto.ts
```diff
@@ -38,36 +53,60 @@ export const formatNewmAmount = (
...
+    precision: 2,
+  }
 ) => {
   if (!amount) return "₳ 0";
 
-  const { includeSymbol = true, precision = 2 } = options;
+  const {
+    includeCurrencySymbol = true,
+    includeEstimateSymbol = false,
+    precision = 2,
```
dmkirshon, line 67: No changes to ADA besides adding the estimated symbol. I don't think we currently have any locations that would need to list "less than .." for ADA values. Precision might need to be evaluated at that point as well. Will bring it up to product to have an eye out.

packages/utils/src/lib/crypto.ts
```diff
@@ -9,27 +9,42 @@ export const NEWM_POLICY_ID = isProd
...
+    precision: 3,
+  }
 ) => {
   if (!amount) return "0 Ɲ";
 
-  const { includeSymbol = true, precision = 2 } = options;
+  const {
+    includeCurrencySymbol = true,
+    includeEstimateSymbol = false,
+    precision = 3,
```
dmkirshon, line 34: Are these deconstruction default values necessary when the params already have default values? Existed before I touched this, but added estimate symbol for deconstruction use and default value of false for consistency.

apps/studio/src/pages/home/library/MarketplaceTab/CreateSale.tsx
```diff
@@ -98,13 +99,25 @@ export const CreateSale = () => {
...
+  const getDynamicDecimalPrecision = (num: number): number | undefined => {
+    if (num >= 0.001 || num === 0 || !Number.isFinite(num)) {
+      return undefined;
+    }
+    const splitDecimalPortion = num.toFixed(10).split(".")[1];
+    const firstNonZeroIndex = splitDecimalPortion.search(/[1-9]/);
+    return firstNonZeroIndex && firstNonZeroIndex !== -1
+      ? firstNonZeroIndex + 1
+      : undefined;
+  };
```
dmkirshon, line 119: Thinking of refactoring this into crypto or another shared repo utility file. Will add a ticket to take care of refactor as other items in PR are more important to release.

apps/studio/src/pages/home/library/MarketplaceTab/CreateSale.tsx
```diff
@@ -134,16 +147,20 @@ export const CreateSale = () => {
...
         }) => {
-          const perTokenPrice = totalSaleValue / tokensToSell;
+          // tokensToSell is "" when the field is empty, converted to 0
+          const perTokenPrice =
+            Number(tokensToSell) !== 0 ? totalSaleValue / tokensToSell : 0;
           const percentageOfUserStreamTokens =
             (tokensToSell / streamTokensInWallet) * 100;
           const formattedPercentageOfUserStreamTokens = `${percentageOfUserStreamTokens.toFixed(
-            2
+            Math.max(2, 7 - tokensToSell.toString().length)
```
dmkirshon, line 156: This seemed like the most logical dynamic handling of the percentage as it covers up to the 1/(max user tokens).

apps/studio/src/pages/home/library/MarketplaceTab/CreateSale.tsx
```diff
@@ -98,13 +99,25 @@ export const CreateSale = () => {
       ),
     totalSaleValue: Yup.number()
       .required("This field is required")
+      .typeError("")
```
dmkirshon, line 102: Was added as there was a non UX friendly validation error outputting when `-` or `.` was entered on its own into the input field. The field is already of type number so other symbols/characters are restricted anyway.

Validation error:
`totalSaleValue must be a `number` type, but the final value was: `NaN` (cast from the value `NaN`).`

apps/marketplace/src/components/Sale.tsx
```diff
@@ -102,7 +103,7 @@ const Sale: FunctionComponent<SaleProps> = ({
 
     return {
       newmAmount: formatNewmAmount(newmAmount),
-      usdAmount: currency(usdAmount).format(),
```
dmkirshon, line 105: Is there a reason why currency was used here instead of  the formatUsdAmount?

escobarjonatan: Great question lol for me, i saw other instances of currency solution so I went with those but if they both achieve the same then I have no problem using the formatXamount method

apps/studio/src/pages/home/library/MarketplaceTab/CreateSale.tsx
escobarjonatan: I created something similarish here: https://github.com/projectNEWM/newm-web/blob/fc53de474cb805982266cc36cd18f08fea75f643/packages/utils/src/lib/sales.ts#L68 so maybe there? but maybe utils package makes more sense?

packages/components/src/lib/earnings/EarningsSummaryModal.tsx
escobarjonatan: The ones i have been adding were because I needed an explicit space and this is one way to achieve that, but with the refactor, if you feel it is not needed, you can remove it just make sure the spacing looks correct in the UI

packages/utils/src/lib/crypto.ts
escobarjonatan: i agree that this should be a part of a different ticket bc it is A LOT.... I feel like our designs went from including 2 decimal spaces, then more then less then more then less lol We should stick to the dynamic one makes sense to stay consistent

packages/utils/src/lib/crypto.ts
escobarjonatan: did you test your theory? lol I THINK we should remove them to decrease redundancy and confusion.

escobarjonatan (approved):

apps/studio/src/pages/home/library/MarketplaceTab/CreateSale.tsx
dmkirshon: Will take a look and create the ticket

packages/components/src/lib/earnings/EarningsSummaryModal.tsx
dmkirshon: Will look at creating a ticket for this one

packages/utils/src/lib/crypto.ts
dmkirshon: Will create the ticket for a review and refactor of currency usage.

apps/marketplace/src/components/Sale.tsx
dmkirshon: I think we just need to do a simple review of them and double check, will be part of the overall review and refactor ticket for currencies across packages.

packages/utils/src/lib/crypto.ts
dmkirshon: Copilot says they are redundant. So it says, so shall it be. 

I believe I quickly tested it, but I would like to spend a few more minutes to double check, smallish but will create a ticket to review this.

PR #822, opened 2024-11-19
Implements flag to handle displaying vespr as wallet option

scandycuz:


apps/marketplace/src/components/header/ConnectWallet.tsx
```diff
@@ -136,7 +138,7 @@ const ConnectWallet: FunctionComponent = () => {
     <Grid>
       <WalletModal
         isOpen={ isConnectWalletModalOpen }
-        omitWallets={ ["vespr"] }
+        omitWallets={ webMarketplaceVesprWallet ? [] : ["vespr"] }
```
escobarjonatan: I haven't tried this option out yet but flags can have different values. One of them being an object so you can go crazy with that as you want

escobarjonatan (approved):

apps/marketplace/src/components/header/ConnectWallet.tsx
scandycuz: Interesting, it could actually be handy to use an array flag so that we can easily disable any wallet. I'll try that approach.

escobarjonatan (approved):

PR #831, opened 2024-11-21
STUD-381: Add feature flag to disable song minting and distribution

escobarjonatan:
image

### DIFF
--- a/apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
+++ b/apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
@@ -11,7 +11,6 @@ import {
   SwitchInputField,
   TextAreaField,
   TextInputField,
-  Tooltip,
   UploadImageField,
   UploadSongField,
 } from "@newm-web/elements";
@@ -131,7 +130,8 @@ const BasicSongDetails: FunctionComponent<BasicDonDetailsProps> = ({
 
   const isSubmitDisabled =
     !values.agreesToCoverArtGuidelines ||
-    (isMintingVisible && (!wallet || !isVerified));
+    (isMintingVisible && (!wallet || !isVerified)) ||
+    (values.isMinting && webStudioDisableTrackDistributionAndMinting);
 
   const handleChangeOwners = (owners: ReadonlyArray<Owner>) => {
     setFieldValue("owners", owners);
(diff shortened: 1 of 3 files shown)


packages/elements/src/lib/SwitchInput.tsx
dmkirshon, line 1: It looks like the `Switch` element is supposed to change to grey/opacity to .7 when disabled, but I'm not sure if we want to also engage disabled view for the icon. Still shows as a white circle when `disabled` currently.

Could open a new ticket for this one, won't block this PR on account to this.

https://github.com/projectNEWM/newm-web/blob/c9dfd65c1baca256ea76f28a992f0ae6c4946f17/packages/elements/src/lib/Switch.tsx#L30-L35

apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
```diff
@@ -401,10 +404,19 @@ const BasicSongDetails: FunctionComponent<BasicDonDetailsProps> = ({
...
                   includeBorder={ false }
                   name="isMinting"
                   title="DISTRIBUTE & MINT SONG"
+                  toggleTooltipText={
+                    webStudioDisableTrackDistributionAndMinting
+                      ? "Track distribution and minting is temporarily disabled. " +
+                        "Please upload your song to save your progress, and check " +
+                        "back later to finish the distribution process."
+                      : undefined
+                  }
```
dmkirshon, line 419: This shows up in the `Library` edit song page when the minting is already toggled by default, I'm able to even proceed to next, logic needs to be added to disable this route as well.

image

dmkirshon (changes requested):
Overall looks good, one minor logic issue spotted on edit page and another optional UI change/inquiry

packages/elements/src/lib/SwitchInput.tsx
escobarjonatan: yea this can be its own ticket

apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
escobarjonatan: good catch, fixed it

dmkirshon: Fix looks good. We will need to make it more obvious though, first instinct isn't to go over to minting toggle to check for tooltip as the reason `Next` button is disabled, might confuse users.

dmkirshon (approved):
Fix looks good to disable minting, one comment left for UX

apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx
escobarjonatan: We can wait on feedback for this as it should just be in for a short(hopefully) period of time

PR #848, opened 2025-01-28
STUD-300: Implement 404 "Page Not Found" error page for NEWM Studio

dmkirshon:
Add:
- Page Not Found component
- Asset for Page Not Found Component

Update:
- Route navigation when attempting to hit a page that doesn't exist

Note:
- Maintained navigation logic for songs and collaboration artist pages, so that if an incorrect ID is used it will fail and return to the appropriate route path.


apps/studio/src/pages/home/uploadSong/UploadSong.tsx
```diff
@@ -139,17 +143,37 @@ const UploadSong: FunctionComponent = () => {
...
-   * TODO: remove this when form values are persisted on refresh
+   * TODO: remove the navigation when form values are persisted on refresh
    */
   useEffect(() => {
-    navigate("/home/upload-song", { replace: true });
+    const validPaths = [
+      "/home/upload-song",
+      "/home/upload-song/advanced-details",
+      "/home/upload-song/confirm",
+    ];
```
scandycuz: This approach could create some issues, as we'll need to make sure to always update this array when adding a new route, or it could result in the 404 component being incorrectly displayed. Instead, it would be better to declare the routes variable externally and then derive this array from them. E.g.

```
const routes = [{
  path: "advanced-details",
  ...
}]

const validRoutes = routes.map((route) =>  `/home/upload-song/${route.path}`)

...

if (!isValidPath) {
  return <PageNotFound redirectUrl="/home/upload-song" />;
}

return (
  <WizardForm
    routes={routes}
    ...
```

That way it only has to be updated in one place, when the routes are declared.

apps/studio/src/pages/createProfile/CreateProfile.tsx
```diff
@@ -59,6 +64,30 @@ const CreateProfile: FunctionComponent = () => {
...
+
+    const normalizePath = (path: string) => path.replace(/\/+$/, ""); // Remove trailing slashes
+    const currentPath = normalizePath(currentPathLocation.pathname);
+
+    setIsValidPath(validPaths.map(normalizePath).includes(currentPath));
+  }, [currentPathLocation.pathname]);
+
+  if (!isValidPath) {
+    return <PageNotFound />;
+  }
```
scandycuz, line 121: Do we need this useEffect? Couldn't we just use
```
if (!isValidPath) {
  return <PageNotFound />;
}
```
directly? It would make the code a bit cleaner.

scandycuz (changes requested):
Looking good but had some comments.

apps/studio/src/pages/home/uploadSong/UploadSong.tsx
scandycuz: The 404 redirect functionality could also be included in the WizardForm component, although I feel less strongly about that.

dmkirshon: Makes sense to consolidate the routes, will make those changes. 

I was trying to keep to the highest level where the routing logic was made. WizardForm is used in multiple different contexts so it felt off to integrate any triggering there.

apps/studio/src/pages/createProfile/CreateProfile.tsx
dmkirshon: Do you mean to remove the useEffect and convert the valid path conditional into a function? I will play around with removing the useEffect and see how the rendering is affected by it.

apps/studio/src/pages/home/uploadSong/UploadSong.tsx
scandycuz: Yeah, I don't have a strong opinion on having the logic in the WizardForm. There are pros and cons to doing it that way.

apps/studio/src/pages/createProfile/CreateProfile.tsx
scandycuz: I supposed a benefit of having it in a `useEffect` call is that it would only run once. It could be fine to keep it there, on second thought.

apps/studio/src/pages/home/uploadSong/UploadSong.tsx
dmkirshon: I refactored the wizard routes and check for valid routes to mimic your example, it should be a little bit cleaner and should reduce future conflicts on updating the routes.

apps/studio/src/pages/createProfile/CreateProfile.tsx
dmkirshon: I removed the useEffect and went for the useMemo route, it cleaned up code and simplified the logic a bit as well. Copilot said there shouldn't be any large rerender performance issues from doing it this way. Let me know if you see any issues with the update on your end.

dmkirshon:
@scandycuz Updated per the comments, some UI issues that were missed, and removed code changes that were committed but were irrelevant to this PR.

scandycuz (approved):
Looks good. The `useMemo` calls work well here, but want to mention that we should be a bit wary about using it and useCallback a lot going forward. Generally, regenerating functions on re-renders is fast enough that `useMemo` and `useCallback` aren't neccessary, and they can add unneeded complexity to the code.

Something else that could be good to work on in a separate PR is also leveraging the 404 page when the ID in a dynamic route is invalid.

PR #849, opened 2025-02-04
MRKT-218: Add token pointer value to sale metadata

escobarjonatan:
image


apps/marketplace/src/components/SaleMetadata.tsx
```diff
@@ -10,6 +10,7 @@ interface SaleMetaDataProps {
 
 interface SaleDetailProps {
   readonly label: string;
+  readonly selectAllText?: boolean;
```
scandycuz: Probably worth changing this to a boolean field name? E.g. something like `isSelectable` or `isSelectableOnClick`?

scandycuz (approved):

PR #859, opened 2025-03-11
STUD-385: Update onboarding and home logic to logout when user becomes unauthorized

escobarjonatan:
For this demo, i added the poll interval on the get profile query in the upload song page to trigger the 401 error of unauthorized. This should mimic closely to what the user sees.

https://github.com/user-attachments/assets/77024c1c-f715-4825-b780-e53da7069b0f

### DIFF
--- a/apps/studio/src/components/OnboardingRedirect.tsx
+++ b/apps/studio/src/components/OnboardingRedirect.tsx
@@ -1,8 +1,6 @@
 import { FunctionComponent, useEffect } from "react";
 import { useNavigate } from "react-router-dom";
-import { FetchBaseQueryError } from "@reduxjs/toolkit/query";
-import { useAppDispatch, useAppSelector } from "../common";
-import { logOutExpiredSession } from "../api/actions";
+import { useAppSelector } from "../common";
 import {
   emptyProfile,
   selectSession,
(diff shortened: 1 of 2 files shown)


apps/studio/src/components/OnboardingRedirect.tsx
```diff
@@ -13,13 +15,27 @@ import {
...
+    // If there’s a 401 error (expired token), exit early.
+    if (error && (error as FetchBaseQueryError).status === 401) {
+      dispatch(logOutExpiredSession());
+      return;
+    }
+
+    // If the user’s profile is incomplete, navigate to the create-profile page.
+    if (!firstName || !lastName || !role) {
       navigate("/create-profile");
     }
```
dmkirshon: As maybe a secondary ticket, should we navigate back to `/home` if the user tries to reach or somehow lands on `/create-profile` when all the details, including location, are already completed?

apps/studio/src/components/OnboardingRedirect.tsx
dmkirshon, line 1: I'm unable to reproduce the issue using your method, not sure what conditions I need to get it to trigger. Maybe I'm doing the polling/trigger set up incorrectly but even if I do what you did in the demo video on the Master branch the functionality is still the same and never actually hits the create-profile page, straight to log out. Were you able to replicate it going to `create-profile` with your specific test steps?

When I use your branch locally or through the preview portal I still see the same issue seen on garage, where if I block the network url for the profile request `me`, or give an empty `{}` on content override, I end up on `/create-profile` on a full page refresh. If I block the url and poll the interval I just end up seeing the error toast message with no navigation. I'm not sure if there is another way to trigger this or its just overall a very rare edge case.

dmkirshon (approved):
LGTM

PR #862, opened 2025-04-03
(TEST) ReferralHero Integration

dmkirshon:


apps/studio/src/App.tsx
```diff
@@ -52,6 +53,25 @@ const App = () => {
...
+      '/'+t+'.js',e.parentNode.insertBefore(c,e)}(window,document,
+      'https://app.referralhero.com/','RHa857ddbc7a');
+    `;
+
+    document.body.appendChild(script);
+
+    return () => {
+      document.body.removeChild(script);
+    };
+  }, []);
```
escobarjonatan, line 73: what do you think about adding this to the index.tsx file like so:

// Function to load the ReferralHero script
const loadReferralHeroScript = () => {
  const script = document.createElement("script");
  script.textContent = "insert string lit here";
  document.body.appendChild(script);
};

// Load the ReferralHero script before rendering the app
loadReferralHeroScript();

escobarjonatan (approved):

scandycuz (approved):

apps/studio/src/App.tsx
dmkirshon: Have no issue implementing it like that, I'm trying to decide what needs to hide behind the LD flag then.

In this case the LD flag won't hide the initial Script but I will then just hide the banners, modals/dashboards, and business logic. I don't think the organic referrals will trigger for tracking on Studio until I switch the target landing page from garage/sign-up to studio/sign-up. So should be good there.

PR #864, opened 2025-04-15
[STUD-425, STUD-426]: Resolves bug when profile updated during song upload

scandycuz:
Prevents upload song page from being reset when profile is fetched.

### Also

- Fixes a minor onboarding styling bug where NEWM logo wasn't centered


dmkirshon (approved):
LGTM, although this looks like it affects the mint toggling after distribution pricing plan is confirmed. Will need to have that fixed in this PR or in a new ticket.

escobarjonatan (approved):

scandycuz:
> LGTM, although this looks like it affects the mint toggling after distribution pricing plan is confirmed. Will need to have that fixed in this PR or in a new ticket.

Resolved the issue.

PR #865, opened 2025-04-15
feat: add referral hero script loading functionality

dmkirshon:
## Add:
- Referral Hero loading script for new campaign


apps/studio/src/referralHeroConfig.ts
```diff
@@ -0,0 +1,17 @@
...
+  const script = document.createElement("script");
+
+  script.textContent = `
+  !function(m,a,i,t,r,e){if(m.RH)return;r=m.RH={},r.uuid
+  =t,r.loaded=0,r.base_url=i,r.queue=[],m.rht=function()
+  {r.queue.push(arguments)};e=a.getElementsByTagName('script')
+  [0],c=a.createElement('script');c.async=!0,c.src=
+  'https://referralhero-global-code.s3.amazonaws.com/'+'production'+
+  '/'+t+'.js',e.parentNode.insertBefore(c,e)}(window,document,
+  'https://app.referralhero.com/','RHa857ddbc7a');
```
dmkirshon: Should this UUID be organized in a .env file or Constant? It isn't a secret key, just a tracker ID, but there might be multiple campaigns. Could be a problem for future integration and refactoring but right now to keep it simple I left it in the script for this initial campaign.

escobarjonatan (approved):

apps/studio/src/referralHeroConfig.ts
dmkirshon, line 1: Updated this so future campaigns can be called

packages/env/src/lib/env.ts
```diff
@@ -47,6 +47,17 @@ const getAppEnvVar = (name: string): string => {
...
+      return typeof process !== "undefined"
+        ? process.env
+            .NEXT_PUBLIC_REFERRALHERO_ARTIST_REFERRAL_CAMPAIGN_UUID_STAGING
+        : import.meta.env
+            .VITE_REFERRALHERO_ARTIST_REFERRAL_CAMPAIGN_UUID_STAGING;
+    case "REFERRALHERO_ARTIST_REFERRAL_CAMPAIGN_UUID_PROD":
+      return typeof process !== "undefined"
+        ? process.env
+            .NEXT_PUBLIC_REFERRALHERO_ARTIST_REFERRAL_CAMPAIGN_UUID_PROD
+        : import.meta.env.VITE_REFERRALHERO_ARTIST_REFERRAL_CAMPAIGN_UUID_PROD;
```
dmkirshon: @scandycuz  I'm just following the pattern here, but is this specific pattern necessary since this won't have a next/marketplace integration? Is there a different pattern that should be followed?

packages/env/src/lib/env.ts
dmkirshon, line 1: Separate issue, but I see that the `process.env.NEXT_PUBLIC` and `import.meta.env.VITE` references are all flagged as type errors since they can "possibly" be undefined. Is there any actual case that they might be undefined or cause issue? Otherwise I will fix it by default `??` the value to empty string `""`

escobarjonatan: Is there a case? Yes if not set lol I think the default makes sense

escobarjonatan (approved):

escobarjonatan (approved):

scandycuz (approved):

PR #868, opened 2025-05-05
STUD-447: end sold out sale

scandycuz:
- Adds flow for ending a sold out sale
- Updates NEWM wallet connector package patch version

### DIFF
--- a/apps/studio/src/components/sales/PingSaleComplete.tsx
+++ b/apps/studio/src/components/sales/PingSaleComplete.tsx
@@ -1,0 +1,126 @@
+import { FunctionComponent, useCallback, useEffect, useState } from "react";
+import { SaleStatus } from "@newm-web/types";
+import {
+  FIFTEEN_SECONDS_IN_MILLISECONDS,
+  FIVE_MINUTES_IN_MILLISECONDS,
+} from "@newm-web/utils";
+import { useGetSalesQuery } from "../../modules/sale";
+import {
+  LOCAL_STORAGE_SALE_COMPLETE_PENDING_KEY,
+  SALE_COMPLETE_UPDATED_EVENT,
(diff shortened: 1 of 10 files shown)


apps/studio/src/components/sales/PingSaleComplete.tsx
```diff
@@ -0,0 +1,126 @@
...
+    {
+      pollingInterval: currentPollingInterval,
+      skip:
+        saleCompleteSongIds === null ||
+        (saleCompleteSongIds && saleCompleteSongIds.length === 0),
+    }
+  );
+
+  const handleSaleComplete = useCallback(() => {
+    const pendingSales = localStorage.getItem(
```
dmkirshon, line 34: Nothing to block here just curious why the use of `localStorage` vs our utils `LocalStorage`

https://github.com/projectNEWM/newm-web/blob/22658d7fcf36045253dcccdb1bbdbbc91c751f74/packages/utils/src/lib/storage.ts#L5

dmkirshon (approved):
LGTM, minor non-blocking comment on util of `localStorage`

apps/studio/src/components/sales/PingSaleComplete.tsx
scandycuz: Most of this code was taken from the [PingSaleEnd](https://github.com/projectNEWM/newm-web/blob/master/apps/studio/src/components/sales/PingSaleEnd.tsx), which used `localStorage`, so it was just from copying and pasting. I agree we should use the util in new code though.

scandycuz: Going to leave it as is for now, just because I've already tested everything and it's a high priority fix. I can update it in a future PR.

PR #876, opened 2025-05-15
STUD-413-Integrate-ReferralHero-modal-dashboard-for-referrer-link-sharing

dmkirshon:
Add:
- Add referral dashboard that can be accessed through the referral banner for the "Artist Referral Campaign"

### DIFF
--- a/apps/studio/src/App.tsx
+++ b/apps/studio/src/App.tsx
@@ -5,7 +5,10 @@ import { Navigate, Route, Routes } from "react-router-dom";
 import theme from "@newm-web/theme";
 import { PersistGate } from "redux-persist/integration/react";
 import { GoogleOAuthProvider } from "@react-oauth/google";
-import { GOOGLE_CLIENT_ID } from "@newm-web/env";
+import {
+  GOOGLE_CLIENT_ID,
+  REFERRALHERO_ARTIST_REFERRAL_CAMPAIGN_UUID,
+} from "@newm-web/env";
 import {
   LDProvider,
   Maintenance,
@@ -25,6 +28,7 @@ import {
   PingSaleStart,
   PrivateRoute,
   ProgressBarModal,
+  ReferralDashboard,
   Toast,
   UpdateWalletAddressModal,
   WalletEnvMismatchModal,
@@ -65,6 +69,9 @@ const App = () => {
                 <IdenfyPingUserStatus />
                 <IdenfyModal />
                 <ConnectWalletModal />
+                <ReferralDashboard
+                  campaignUUID={ REFERRALHERO_ARTIST_REFERRAL_CAMPAIGN_UUID }
+                />
                 <InvitesModal />
                 <ProgressBarModal />
                 <UpdateWalletAddressModal />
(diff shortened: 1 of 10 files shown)


apps/studio/src/App.tsx
```diff
@@ -65,6 +69,9 @@ const App = () => {
                 <IdenfyPingUserStatus />
                 <IdenfyModal />
                 <ConnectWalletModal />
+                <ReferralDashboard
+                  campaignUUID={ REFERRALHERO_ARTIST_REFERRAL_CAMPAIGN_UUID }
+                />
```
dmkirshon, line 74: This modal is only visible via interaction with the Referral Banner, should I still put it behind the LD flag?

apps/studio/src/components/referralHero/ReferralDashboard.tsx
```diff
@@ -0,0 +1,287 @@
...
+/**
+ * ReferralDashboard component displays a modal with referral stats and sharable link.
+ * If the user doesn't have a wallet connected, it shows a wallet connection overlay.
+ */
+const ReferralDashboard: FunctionComponent<ReferralDashboardProps> = ({
+  campaignUUID,
+}) => {
+  const { webStudioArtistReferralCampaign } = useFlags();
+  // Check if the referral campaign is enabled before rendering the wrapper component
+  if (!webStudioArtistReferralCampaign) return null;
```
dmkirshon, line 49: Added this to stop the component from loading if the flag is false as this is a wrapper and would load even if the campaign is not active. Probably not what we want.

apps/studio/src/components/referralHero/ReferralDashboard.tsx
```diff
@@ -0,0 +1,287 @@
...
+  const theme = useTheme();
+  const dispatch = useAppDispatch();
+  const { isLoggedIn } = useAppSelector(selectSession);
+  const { isReferralDashboardModalOpen } = useAppSelector(selectUi);
+
+  // Saved wallet address will be used for referral rewards
+  const { data: { walletAddress } = emptyProfile } = useGetProfileQuery(
+    undefined,
+    { skip: !isLoggedIn }
+  );
```
dmkirshon, line 60: Switched from `wallet` connected conditional to `walletAddress`  conditional so that once a user already has a wallet address saved in their user profile there is no need to show the connect wallet dashboard blocker. 

Not sure if we want a dialog or something to say that the reward will go to the attached wallet addressed... will bring up to product.

apps/studio/src/App.tsx
dmkirshon: I added a LD flag within the component to prevent loading it for all users if the flag is false.

packages/components/src/lib/Alert.tsx
```diff
@@ -29,10 +29,12 @@ const Alert: FunctionComponent<AlertProps> = ({
     setIsOpen(false);
   };
 
-  return isOpen ? (
+  // Without message check the alert is set to empty but the snackbar
+  // persists for a short period when clearToastMessage is called
+  return isOpen && !!message ? (
```
dmkirshon, line 34: I added this as there were some points where a success message, for the link copy, would quickly be followed by an empty error message due to the `clearToastMessage`. From `message` being a required attribute, the assumption is that `message` will always be available to show within the `Snackbar`. If this assumption is incorrect and an empty `""` `message` is a possibility then I will revert this change.

escobarjonatan (approved):

PR #877, opened 2025-05-16
STUD-459: Add referrer support to sign-up flows

dmkirshon:
Enhance sign-up flows by adding support for a referrer parameter, allowing tracking of referral sources during user registration and login processes. This update modifies the login components and API requests to include the referrer information.

### DIFF
--- a/apps/studio/src/modules/session/api.ts
+++ b/apps/studio/src/modules/session/api.ts
@@ -62,11 +62,16 @@ export const extendedApi = newmApi.injectEndpoints({
         }
       },
 
-      query: (body) => ({
-        body,
-        method: "POST",
-        url: "v1/auth/login/apple",
-      }),
+      query: (body) => {
+        const { referrer, ...bodyData } = body;
+
+        return {
+          body: bodyData,
+          method: "POST",
+          params: referrer ? { referrer: referrer } : undefined,
+          url: "v1/auth/login/apple",
+        };
+      },
     }),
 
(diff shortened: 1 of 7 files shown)


apps/studio/src/modules/session/api.ts
```diff
@@ -104,11 +109,16 @@ export const extendedApi = newmApi.injectEndpoints({
         }
       },
 
-      query: (body) => ({
-        body,
-        method: "POST",
-        url: "v1/users",
-      }),
+      query: (body) => {
+        const { referrer, ...bodyData } = body;
```
dmkirshon, line 113: Does it make more sense to split this into params and body?

apps/studio/src/pages/signUp/SignUp.tsx
```diff
@@ -18,18 +18,24 @@ interface AccountValues {
...
 }
 
 const SignUp: FunctionComponent = () => {
   const theme = useTheme();
   const dispatch = useAppDispatch();
   const currentPathLocation = useLocation();
 
+  // Parse query parameters for referral code
+  const queryParams = new URLSearchParams(currentPathLocation.search);
+  const referrerCode = queryParams.get("mwr") || "";
```
dmkirshon, line 31: Should this also hide behind the LD flag, I don't believe it will do anything at this point without the Studio production campaign is created.

escobarjonatan (approved):

PR #880, opened 2025-06-16
STUD-353: updates wallet connector version

scandycuz:
Updates wallet connector version with wallet duplicate signature fix. Confirmed fix works locally.


dmkirshon (approved):
LGTM, tested locally as well with no issues.

PR #892, opened 2025-06-23
STUD-471: Fix error state of songs when edit flow payment tx errors occur

dmkirshon:
Add additional minting statuses to allow users to edit songs
even after payment issues, improving user experience during
the minting process.

### DIFF
--- a/apps/studio/src/common/song.ts
+++ b/apps/studio/src/common/song.ts
@@ -3,6 +3,9 @@ import { MintingStatus } from "@newm-web/types";
 export const isSongEditable = (mintingStatus: MintingStatus) => {
   return (
     mintingStatus === MintingStatus.Undistributed ||
-    mintingStatus === MintingStatus.Declined
+    mintingStatus === MintingStatus.Declined ||
+    // Keeps user on edit flow even after payments fail after a tx error
+    mintingStatus === MintingStatus.MintingPaymentRequested ||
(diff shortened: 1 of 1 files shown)


apps/studio/src/common/song.ts
dmkirshon, line 1: There is probably more complicated logic that would need to be update to make the process more succinct. I believe this at least allows the user to be unblocked from submitting the payment transaction when the initial signature transaction fails due to a wallet or user caused issue in the "Edit Song" flow. 

I also noticed the following that hopefully can get incorporated into the future UX updates:
1. Even though the song data was picking up in a minting status of `MintingPaymentRequested`, the page was still acting on cached data with a minting status of `StreamTokenAgreementApproved`, this was also causing navigation to the `view-details` page, ending any ability for the user to select "Confirm & Pay". Without including both statuses to the editable list it would have required a page refresh for the user to get back to edit mode.
2. There is still no UI or UX to tell the user to retry payment or information that there are issues with some wallets:
   - With the "Upload a Song" flow the upload progress bar disappears and that shows some indication that there was failure (which may or may not lead them to select "Confirm & Pay"  again)
   - With the "Edit Song" flow the "Confirm & Pay" button, on click, doesn't show loading or any sign that a transaction is in progress. On tx failure, a toast message does appear, but that gives no hint toward pressing the "Confirm & Pay" button again to try and engage the process again.

scandycuz (approved):

PR #901, opened 2025-08-01
MRKT-113-404-Page-Not-Found-error-page-for-Pages-Sales-and-Artist-in-Marketplace

dmkirshon:
Introduce a NotFound component to improve user experience by redirecting users to a 404 page when an artist or sale is not found. Adjust visual and responsive consistency of the PageNotFound component. Refine error handling in queries to no longer show error toast messages for artist and song sale query leading to 404 page.

[MRKT-113 Artist Page Not Found.webm](https://github.com/user-attachments/assets/f6cbad2e-9a82-4560-a6c8-601e340e5030)

[MRKT-113 Song Sale Page Not Found.webm](https://github.com/user-attachments/assets/897a9ae5-af21-4c5c-8705-bae078ee7a9e)

### DIFF
--- a/apps/marketplace/src/modules/artist/api.ts
+++ b/apps/marketplace/src/modules/artist/api.ts
@@ -12,19 +12,6 @@ import { Tags } from "../../api/newm/types";
 export const extendedApi = newmApi.injectEndpoints({
   endpoints: (build) => ({
     getArtist: build.query<GetArtistResponse, string>({
-      async onQueryStarted(body, { dispatch, queryFulfilled }) {
-        try {
-          await queryFulfilled;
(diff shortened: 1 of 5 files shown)


apps/marketplace/src/modules/artist/api.ts
dmkirshon, line 1: I removed the Toast Error Message as instructed by Product, as the default is now for the Page Not Found to be navigated to. Though, I also asked if they would still like us to show the Toast Error Message when other Errors, other than 404s, are reached. It was a generic message before but we could at least filter out the 404 error situation.

escobarjonatan (approved):

PR #912, opened 2025-10-17
STUD-503: Fix incorrect Release date in enhanced Payment Summary 

dmkirshon:
Update:
- Introduce a utility function to update ISO dates as UTC dates and convert into locale-specific format. 
- Update moment.js integration to utilize browser locale settings.
- Add unit tests for the time utility functions and the release summary dialog.

[STUD-503](https://projectnewm.atlassian.net/browse/STUD-503)

[STUD-503]: https://projectnewm.atlassian.net/browse/STUD-503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/studio/src/index.tsx
+++ b/apps/studio/src/index.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import ReactDOM from "react-dom/client";
 import "global.css";
+import moment from "moment/min/moment-with-locales";
 import App from "./App";
 import reportWebVitals from "./reportWebVitals";
 import initializeSentry from "./sentryConfig";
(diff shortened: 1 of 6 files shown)


apps/studio/src/index.tsx
```diff
@@ -14,6 +15,9 @@ if (window.location.hostname.includes("newm-studio")) {
 initializeSentry();
 loadReferralHeroScript();
 
+const browserLocale = navigator.language.toLowerCase();
+moment.locale(browserLocale);
```
dmkirshon, line 19: Properly formats moment based calls to use locale formatting. the library moment.js does not automatically grab the browser's locale. The default was 'en' which ends up always displaying United States formatting rules, for example when `date.format("L")` is used. 

Moving forward instead of importing as `import moment from "moment";` we should instead use `import moment from "moment/min/moment-with-locales";`. Testing should also handle multiple locale options to confirm date formats appear as they should.

escobarjonatan (approved):

PR #913, opened 2025-10-28
STUD-512: displays error message when Cloudinary upload fails

scandycuz:
Updates the Cloudinary API upload image endpoint to display an toast with an error message if it fails.

Screen Shot 2025-10-28 at 2 28 51 AM


dmkirshon (approved):
LGTM

PR #933, opened 2026-01-14
[STUD-539][enhancement]: Add "Create New Release" button to the Releases page header

ITurres:
# Pull Request - Issue #[STUD-539](https://projectnewm.atlassian.net/browse/STUD-539)

## Overview

> This PR adds a prominent "Create New Release" CTA tile to the Studio Releases (Discography.tsx) page and introduces a reusable gradient-dashed outline primitive for consistent styling.
> Context: we needed a dashed border with a NEWM gradient treatment; `border-image` overrides dashed borders and theme gradients cannot be applied via `border-color`, so a dedicated component was created using layered backgrounds to achieve the desired effect.

---

## Files Summary

> **Total Changes:** 4 files  
> (Counts of added, modified, and deleted files can be seen in the PR diff view.)

<details>
<summary>Added (2)</summary>

- **`packages/elements/src/lib/styled/GradientDashedOutline.tsx`**
  - Styled `Box` primitive that renders a dashed outline with a configurable theme gradient using layered backgrounds.
- **`packages/elements/src/lib/styled/GradientDashedOutline.test.tsx`**
  - Unit tests covering render behavior, prop filtering, and gradient-driven class updates.

</details>

<details>
<summary>Modified (2)</summary>

- **`packages/elements/src/index.ts`**
  - Exports `GradientDashedOutline` from `@newm-web/elements`.
- **`apps/studio/src/pages/home/releases/Discography.tsx`**
  - Adds a clickable "Create New Release" CTA tile linking to `/home/upload-song`.

</details>

<details>
<summary>Deleted (0)</summary>

</details>

---

## Impact

- Improves discoverability of the release creation/upload flow by adding a clear CTA in Discography.
- Introduces a reusable `GradientDashedOutline` primitive; existing `DashedOutline` usage remains unchanged (backwards compatible).
- No runtime dependency changes.

---

## Testing

- Unit tests:
  - `npx nx test elements` or `npx nx test elements --testPathPattern=GradientDashedOutline.test.tsx`
- Manual:
  - Verify Discography shows "Create New Release" tile and clicking it navigates to `/home/upload-song`.

---

## Related Issues

- Closes #[STUD-539](https://projectnewm.atlassian.net/browse/STUD-539)

---

## Dependencies

- No new dependencies.

---

## Demo

https://github.com/user-attachments/assets/9b0e25f2-06e4-4915-9b9c-5db4fb572550

---

## Additional Notes

- Why a new component was needed:
  - `border-image`/`borderImage` paints over the border and does not preserve a `dashed` stroke.
  - Theme gradients are `linear-gradient(...)` strings and cannot be applied via `borderColor`.
  - `GradientDashedOutline` uses a gray-700 dashed border + layered background clips (`padding-box` / `border-box`) to create a "gradient dashed" outline effect without requiring SVG.

--END--

[STUD-539]: https://projectnewm.atlassian.net/browse/STUD-539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[STUD-539]: https://projectnewm.atlassian.net/browse/STUD-539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/studio/src/pages/home/releases/Discography.tsx
+++ b/apps/studio/src/pages/home/releases/Discography.tsx
@@ -1,9 +1,8 @@
 import { FunctionComponent, useState } from "react";
-import { Link } from "react-router-dom";
 
 import { Box, Typography } from "@mui/material";
 
-import { GradientDashedOutline, IconMessage } from "@newm-web/elements";
+import { GradientDashedOutline, IconMessage, Link } from "@newm-web/elements";
 import { AddSong } from "@newm-web/assets";
 
 import SongList from "./SongList";
@@ -29,7 +28,13 @@ const Discography: FunctionComponent = () => {
       </Typography>
 
       <Box sx={ { mb: 5.5 } }>
-        <Link aria-label="Create New Release" to="/home/upload-song">
+        <Link
+          aria-label="Create New Release"
+          sx={ {
+            textDecoration: "none",
+          } }
+          to="/home/upload-song"
+        >
           <GradientDashedOutline
             sx={ {
               padding: 3,
--- a/apps/studio/src/pages/home/releases/Discography.tsx
+++ b/apps/studio/src/pages/home/releases/Discography.tsx
@@ -1,7 +1,7 @@
 import { FunctionComponent, useState } from "react";
 import { Link } from "react-router-dom";
 
-import { Typography } from "@mui/material";
+import { Box, Typography } from "@mui/material";
 
 import { GradientDashedOutline, IconMessage } from "@newm-web/elements";
 import { AddSong } from "@newm-web/assets";
@@ -28,16 +28,17 @@ const Discography: FunctionComponent = () => {
         RELEASES
       </Typography>
 
-      <Link aria-label="Create New Release" to="/home/upload-song">
-        <GradientDashedOutline
-          sx={ {
-            marginBottom: 5.5,
-            padding: 3,
-          } }
-        >
-          <IconMessage icon={ <AddSong /> } message="Create New Release" />
-        </GradientDashedOutline>
-      </Link>
+      <Box sx={ { mb: 5.5 } }>
+        <Link aria-label="Create New Release" to="/home/upload-song">
+          <GradientDashedOutline
+            sx={ {
+              padding: 3,
+            } }
+          >
+            <IconMessage icon={ <AddSong /> } message="Create New Release" />
+          </GradientDashedOutline>
+        </Link>
+      </Box>
 
       { totalCountOfSongs || query ? (
         <SearchBox
--- a/apps/studio/src/pages/home/releases/Discography.tsx
+++ b/apps/studio/src/pages/home/releases/Discography.tsx
@@ -1,5 +1,11 @@
 import { FunctionComponent, useState } from "react";
+import { Link } from "react-router-dom";
+
 import { Typography } from "@mui/material";
+
+import { GradientDashedOutline, IconMessage } from "@newm-web/elements";
+import { AddSong } from "@newm-web/assets";
+
 import SongList from "./SongList";
 import { SearchBox } from "../../../components";
 import { useGetSongCountQuery } from "../../../modules/song";
@@ -21,6 +27,17 @@ const Discography: FunctionComponent = () => {
       <Typography sx={ { pb: 4 } } variant="h3">
         RELEASES
       </Typography>
+
+      <Link aria-label="Create New Release" to="/home/upload-song">
+        <GradientDashedOutline
+          sx={ {
+            marginBottom: 5.5,
+            padding: 3,
+          } }
+        >
+          <IconMessage icon={ <AddSong /> } message="Create New Release" />
+        </GradientDashedOutline>
+      </Link>
 
       { totalCountOfSongs || query ? (
         <SearchBox
--- a/packages/elements/src/lib/styled/GradientDashedOutline.tsx
+++ b/packages/elements/src/lib/styled/GradientDashedOutline.tsx
@@ -1,0 +1,32 @@
+import { Box } from "@mui/material";
+import { Theme, styled } from "@mui/material/styles";
+import theme from "@newm-web/theme";
+
+type GradientKey = keyof Theme["gradients"];
+
+interface GradientDashedOutlineProps {
+  gradient?: GradientKey;
+}
+
+/**
+ * * Dashed outline with a configurable theme gradient border.
+ *
+ * * Uses a grey700 dashed border + layered backgrounds to create a
+ * * "gradient dashed border" effect. Prefer this for CTA tiles (e.g. "Create new")
+ * * where a gradient border is desired without introducing SVG.
+ */
+const GradientDashedOutline = styled(Box, {
+  shouldForwardProp: (prop) => prop !== "gradient",
+})<GradientDashedOutlineProps>(({ gradient = "newm" }) => ({
+  "&:hover": {
+    backgroundImage: `linear-gradient(${theme.colors.grey600}, ${theme.colors.grey600}), ${theme.gradients[gradient]}`,
+  },
+  backgroundClip: "padding-box, border-box",
+
+  backgroundImage: `linear-gradient(${theme.colors.grey700}, ${theme.colors.grey700}), ${theme.gradients[gradient]}`,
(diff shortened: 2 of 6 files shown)


ITurres:
I'll move the `GradientDashedOutline.test.tsx` to the dedicated `test` dir.

apps/studio/src/pages/home/releases/Discography.tsx
```diff
@@ -22,6 +28,17 @@ const Discography: FunctionComponent = () => {
...
+      <Link aria-label="Create New Release" to="/home/upload-song">
+        <GradientDashedOutline
+          sx={ {
+            marginBottom: 5.5,
+            padding: 3,
+          } }
+        >
+          <IconMessage icon={ <AddSong /> } message="Create New Release" />
+        </GradientDashedOutline>
+      </Link>
```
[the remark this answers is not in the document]
ITurres: Cursor already shows as pointer on hover since Link renders an `<a>` and `GradientDashedOutline` doesn't override cursor, so no explicit `cursor: pointer` needed. However, I found a 'pointer on whitespace' issue at the bottom spacing of `GradientDashedOutline`; moving the margin outside the clickable tile fixed that.

packages/elements/src/lib/styled/GradientDashedOutline.tsx
```diff
@@ -0,0 +1,32 @@
...
+  shouldForwardProp: (prop) => prop !== "gradient",
+})<GradientDashedOutlineProps>(({ gradient = "newm" }) => ({
+  "&:hover": {
+    backgroundImage: `linear-gradient(${theme.colors.grey600}, ${theme.colors.grey600}), ${theme.gradients[gradient]}`,
+  },
+  backgroundClip: "padding-box, border-box",
+
+  backgroundImage: `linear-gradient(${theme.colors.grey700}, ${theme.colors.grey700}), ${theme.gradients[gradient]}`,
+  backgroundOrigin: "padding-box, border-box",
+  border: `2px dashed ${theme.colors.grey700}`,
```
dmkirshon, line 28: I think this was a good simple implementation. The color scheme of the gradient does look different from the figma though. The strong yellow at the left and color pattern difference seem a little overpowering from my side, but please follow up with Denny/Design to make sure the gradient orientation is still ok like this.

Otherwise I have no issue releasing this way currently. If we need to adapt it to meet the Figma spec with an SVG component and some new Theme parameters for Gradient borders, I have no issue with that either. We have run into the gradient issues before and had to make some custom changes for certain components before.

apps/studio/src/pages/home/releases/Discography.tsx
```diff
@@ -1,5 +1,11 @@
 import { FunctionComponent, useState } from "react";
-import { Typography } from "@mui/material";
+import { Link } from "react-router-dom";
```
dmkirshon: I believe we have our own `Link` element that uses `react-router-dom` but also integrates MUI typography. Does it make sense to utilize that element here or is that element not compatible with the set up? 


Note: This is another area to watch out for, where I think we have used the `Link` component from `@mui/material`, should only be used in for external links, in some places where it shouldn't. This is internal so we are good. ~~I will create a ticket that captures this as I think our maintenance and page-not-found use the incorrect `Link` after looking it over.~~ Nevermind on me creating this ticket, I think everything is currently referenced correctly.

apps/studio/src/pages/home/releases/Discography.tsx
```diff
@@ -22,6 +28,18 @@ const Discography: FunctionComponent = () => {
         RELEASES
       </Typography>
 
+      <Box sx={ { mb: 5.5 } }>
+        <Link aria-label="Create New Release" to="/home/upload-song">
```
dmkirshon: Not blocking for this PR but please validate with product if we need to keep the `/home/upload-song` path or if we need to have it restructured to `/home/releases/upload-song` for this Phase 1 Epic. We can integrate the change, if needed, as part of the SideNavBar and Library clean up.

dmkirshon (approved):
Approved with no major issues. Three comments, but none are blocking potential. Please review and address where needed. If no major shifts are identified, comment and push forward.

apps/studio/src/pages/home/releases/Discography.tsx
ITurres: On it @dmkirshon

packages/elements/src/lib/styled/GradientDashedOutline.tsx
ITurres: It does look different, Ok Checking 💪... Thanks david!

ITurres: Gradient values updated to closely match the Figma spec! 🚀 

image (16)

apps/studio/src/pages/home/releases/Discography.tsx
ITurres: Yes we do @dmkirshon, at `newm-web/packages/elements/src/lib/Link.tsx`. The only visual override it adds is `text-decoration: underline`, which makes the CTA differ from the Figma design (see attached).

image

ITurres: Style that we can override when using our `Link` at `Discography.tsx`.

Overall it's preference: native router `Link` matches Figma by default. using our wrapper just means overriding `underline`. Since we're using our wrapper today, the override seems fine. Thoughts?

apps/studio/src/pages/home/releases/Discography.tsx
ITurres: Linking for context https://projectnewm.atlassian.net/browse/STUD-538?focusedCommentId=21032, so we decide

apps/studio/src/pages/home/releases/Discography.tsx
dmkirshon: My take is to override the `underline` and use the element wrapper in this case. The Link element makes the most sense to use consistently across our codebase right now. 

We will need to review the `SideBarNavLink.tsx` file as that is the other place we use the `react-router-dom` reference. I added it as a Dev Note to [STUD-538](https://projectnewm.atlassian.net/browse/STUD-538), but this can be moved to a smaller chore ticket.

If you've run into these similar components being used across other codebases, please let us know if you encountered a better way to handle the differences/imports.

packages/theme/src/lib/theme.ts
```diff
@@ -212,7 +212,7 @@ const theme = createTheme({
     magazine: "linear-gradient(53.48deg, #F53C69 0%, #FF6E32 100%)",
     music: "linear-gradient(53.48deg, #C341F0 0%, #F53C69 100%)",
     // eslint-disable-next-line max-len
-    newm: "linear-gradient(45.38deg, #FFC33C 14.22%, #FF6E32 28.39%, #F53C69 42.57%, #C341F0 56.74%, #5091EB 70.91%, #41BE91 85.83%)",
+    newm: "linear-gradient(12deg, #FFC33C 0%, #FF6E32 24%, #F53C69 41%, #C341F0 59%, #5091EB 70%, #41BE91 90%)",
```
dmkirshon, line 219: I think a comment should be left here to note that this deviates from the figma "newm" gradient, but that it meets the same intent as the design pattern `var(--Gradients-Gradient-NEWM-45);`  (Gradients/Gradient - NEWM 45°)

I couldn't find any other reference to using `newm` gradient across our codebase, but just so the next developer understands if the pattern doesn't end up matching the figma design in another component design.

ITurres: Agree! Comment added. And yeah, the `newm` gradient isn't referenced anywhere else at the moment, nor accessed dynamically, except in `GradientDashedOutline.tsx` for the CTA.

apps/studio/src/pages/home/releases/Discography.tsx
ITurres: Done! File updated. And thanks for adding that dev note, standardizing Link usage (incl. `SideBarNavLink`) will definitely help longer term.

PR #936, opened 2026-01-19
[STUD-541][Enhancement]: Add three dot action menu and conditional delete release functionality

ITurres:
# Pull Request — Issue [STUD-541](https://projectnewm.atlassian.net/browse/STUD-541)

## Overview

> Adds a three-dot action menu to Releases rows with View/Edit and conditional Delete, including confirmation modal, tooltip gating, and localStorage cleanup.  
> Also introduces reusable ActionMenu/ActionMenuTrigger components (with tests) to standardize action menus across the app.
>
> **Other tickets minor changes:**
>
> - `GradientDashedOutline` style updates. Refer to [STUD-539-comments](https://projectnewm.atlassian.net/browse/STUD-539?focusedCommentId=21120)
> - `TableHead` copy updates. Refer to [STUD-540-comments](https://projectnewm.atlassian.net/browse/STUD-540?focusedCommentId=21108)
> - `EditSong` tooltip only renders when non-deletable. Refer to [STUD-542-comments](https://projectnewm.atlassian.net/browse/STUD-542?focusedCommentId=21111)

---

## Files Summary

> **Total Changes:** 17 files  
> (Counts of added, modified, and deleted files can be seen in the PR diff view.)

<details>
<summary>Added (7)</summary>

- **`packages/components/src/lib/ActionMenu.tsx`**
  - — Reusable action menu with configurable items, tooltips, dividers, and anchors.
- **`packages/components/src/lib/ActionMenuTrigger.tsx`**
  - — Standardized kebab trigger button for action menus.
- **`packages/components/src/lib/test/ActionMenu.test.tsx`**
  - — Unit tests for ActionMenu behaviors.
- **`packages/components/src/lib/test/ActionMenuTrigger.test.tsx`**
  - — Unit tests for ActionMenuTrigger.
-**`packages/utils/src/lib/storage.test.ts`.
   - — Unit tests for all storage-related functions. 

  - **`ActionMenuTrigger` is kept separate from `ActionMenu` so the menu remains anchor‑agnostic and reusable; different screens can supply their own trigger (kebab, text button, icon) without coupling layout/trigger styling to the menu component.**

- **`packages/components/src/lib/test/utils/render.tsx`**
  - — Shared test renderer with theme provider.
- **`packages/components/src/setupTests.js`**
  - — Jest setup with env mocks for components package.

</details>

<details>
<summary>Modified (10)</summary>

- **`apps/studio/src/pages/home/releases/SongList.tsx`**
  - — Adds row action menu, delete flow (archived:true), and pending-sale localStorage cleanup.
- **`apps/studio/src/pages/home/releases/EditSong.tsx`**
  - — Delete tooltip only renders when non-deletable.
- **`apps/studio/src/pages/home/releases/DeleteSongModal.tsx`**
  - — Copy and styling tweaks for deletion confirmation.
- **`apps/studio/src/pages/home/releases/Table/TableHead.tsx`**
  - — Header copy updates and action column alignment.
- **`apps/studio/src/pages/home/releases/Discography.tsx`**
  - — Search placeholder/copy updates.
- **`packages/components/jest.config.ts`**
  - — Adds setupTests for components test environment.
- **`packages/components/src/index.ts`**
  - — Exports new ActionMenu/Trigger components.
- **`packages/elements/src/lib/styled/GradientDashedOutline.tsx`**
  - — Style updates to match refreshed UI.
-**`packages/utils/src/lib/storage.ts`.
   - — Added new `parseStoredJSON` func.
-**`packages/utils/src/lib/storage.ts`.
   - — Added a backward‑compatible new `deleteSong` argument variant to optionally redirect to /home/releases while keeping existing callers unchanged.

</details>

<details>
<summary>Deleted (0)</summary>

- _None._

</details>

---

## Impact

- Releases table now uses a consistent action menu for View/Edit and conditional Delete.
- Delete action is gated by release status and shows tooltip only when non-deletable.
- New reusable components standardize menu styling and behavior across the app.
- LocalStorage cleanup removes stale pending-sale entries upon delete confirm.

---

## Testing

- Run:
  - `npx nx test components --testPathPattern=ActionMenu.test.tsx`
  - `npx nx test components --testPathPattern=ActionMenuTrigger.test.tsx`
  - `npx nx test utils --testPathPattern=storage.test.ts`

---

## Related Issues

- Closes [STUD-541](https://projectnewm.atlassian.net/browse/STUD-541)
- Addresses minor ticket comments: [STUD-539-comments](https://projectnewm.atlassian.net/browse/STUD-539?focusedCommentId=21120), [STUD-540-comments](https://projectnewm.atlassian.net/browse/STUD-540?focusedCommentId=21108) and [STUD-542-comments](https://projectnewm.atlassian.net/browse/STUD-542?focusedCommentId=21111).

---

## Dependencies

- No new dependencies.

---

## Demo

STUD-541-delete-tooltip

_Note: The delete tooltip uses the reusable `ReleaseDeletionHelp` component, so its `width/line‑wrap` differs slightly from Figma._

https://github.com/user-attachments/assets/b2decebc-b21a-4d5d-9345-86fe005f0101

_Note: Edit/Visibility icons now have correct size [8aa7a1](https://github.com/projectNEWM/newm-web/pull/936/commits/8aa7a1436134ad3aaaa2556bc2c4fedd3cdabec2)._

---

## Additional Notes

- Delete action uses `archived: true` via the existing `deleteSong` endpoint.
- Tooltip for Delete only renders when the status is not deletable.
- The localStorage cleanup repeats similar parse/remove logic for array vs map; if this expands, we should consider extracting a shared utility.

--END--

[STUD-541]: https://projectnewm.atlassian.net/browse/STUD-541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[STUD-539]: https://projectnewm.atlassian.net/browse/STUD-539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[STUD-541]: https://projectnewm.atlassian.net/browse/STUD-541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/studio/src/modules/song/thunks.ts
+++ b/apps/studio/src/modules/song/thunks.ts
@@ -562,17 +562,26 @@ export const getUserWalletSongs = createAsyncThunk(
  * Request to delete user song. If successful, navigate to
  * library and fetch new songs.
  */
+type DeleteSongRedirectVariant = {
+  redirectToReleases?: boolean;
+  request: DeleteSongRequest;
+};
+
 export const deleteSong = createAsyncThunk(
   "song/deleteSong",
-  async (body: DeleteSongRequest, { dispatch }) => {
+  async (body: DeleteSongRequest | DeleteSongRedirectVariant, { dispatch }) => {
     try {
-      // navigate to library page before deleting due to known
+      const request = "request" in body ? body.request : body;
+      const redirectToReleases =
+        "redirectToReleases" in body ? body.redirectToReleases : false;
+
+      // navigate before deleting due to known
       // issue where RTK Query hook will re-fetch data after
       // delete call invalidates cache tag, causing 404 error to
       // display: https://github.com/reduxjs/redux-toolkit/issues/1672
-      history.replace("/home/library");
+      history.replace(redirectToReleases ? "/home/releases" : "/home/library");
 
-      await dispatch(songApi.endpoints.deleteSong.initiate(body));
+      await dispatch(songApi.endpoints.deleteSong.initiate(request));
     } catch (err) {
       // do nothing, errors handled by endpoints
     }
--- a/apps/studio/src/modules/song/thunks.ts
+++ b/apps/studio/src/modules/song/thunks.ts
@@ -560,8 +560,9 @@ export const getUserWalletSongs = createAsyncThunk(
 
 /**
  * Request to delete user song. If successful, navigate to
- * library and fetch new songs.
+ * library / releases (flag variant) and fetch new songs.
  */
+// TODO(webStudioAlbumPhaseOne): Remove redirectToReleases variant once flag is retired.
 type DeleteSongRedirectVariant = {
   redirectToReleases?: boolean;
   request: DeleteSongRequest;
--- a/apps/studio/src/pages/home/releases/EditSong.tsx
+++ b/apps/studio/src/pages/home/releases/EditSong.tsx
@@ -42,6 +42,7 @@ interface EditSongFormValues extends PatchSongThunkRequest {
 }
 
 const EditSong: FunctionComponent = () => {
+  // TODO(webStudioAlbumPhaseOne): Remove flag once flag is retired.
   const { webStudioAlbumPhaseOne } = useFlags();
   const { webStudioDisableTrackDistributionAndMinting } = useFlags();
 
--- a/apps/studio/src/pages/home/releases/EditSong.tsx
+++ b/apps/studio/src/pages/home/releases/EditSong.tsx
@@ -104,6 +104,8 @@ const EditSong: FunctionComponent = () => {
     isLoading: isGetSongLoading,
   } = useGetSongQuery(songId);
 
+  const isSongDeletable = getIsSongDeletable(mintingStatus);
+
   const isDeclined = mintingStatus === MintingStatus.Declined;
 
   const owners = collaborations
@@ -312,11 +314,17 @@ const EditSong: FunctionComponent = () => {
         { title && <Typography variant="h3">{ title.toUpperCase() }</Typography> }
 
         <>
-          <Tooltip title={ <ReleaseDeletionHelp /> }>
+          <Tooltip
+            disableFocusListener={ isSongDeletable }
+            disableHoverListener={ isSongDeletable }
+            disableTouchListener={ isSongDeletable }
+            title={ isSongDeletable ? "" : <ReleaseDeletionHelp /> }
+          >
             <Stack ml="auto">
               <Button
                 color="white"
-                disabled={ !getIsSongDeletable(mintingStatus) }
+                // render other status for
+                disabled={ !isSongDeletable }
                 sx={ { marginLeft: "auto" } }
                 variant="outlined"
                 width="icon"
--- a/apps/studio/src/pages/home/releases/EditSong.tsx
+++ b/apps/studio/src/pages/home/releases/EditSong.tsx
@@ -42,11 +42,12 @@ interface EditSongFormValues extends PatchSongThunkRequest {
 }
 
 const EditSong: FunctionComponent = () => {
+  const { webStudioAlbumPhaseOne } = useFlags();
+  const { webStudioDisableTrackDistributionAndMinting } = useFlags();
+
   const navigate = useNavigate();
   const dispatch = useDispatch();
   const { songId } = useParams<"songId">() as SongRouteParams;
-
-  const { webStudioDisableTrackDistributionAndMinting } = useFlags();
 
   const { data: genres = [] } = useGetGenresQuery();
   const { data: roles = [] } = useGetRolesQuery();
@@ -340,7 +341,10 @@ const EditSong: FunctionComponent = () => {
           { isDeleteModalActive && (
             <DeleteSongModal
               primaryAction={ () => {
-                deleteSong({ songId });
+                deleteSong({
+                  redirectToReleases: Boolean(webStudioAlbumPhaseOne),
+                  request: { songId },
+                });
               } }
               secondaryAction={ () => {
                 setIsDeleteModalActive(false);
--- a/apps/studio/src/pages/home/releases/EditSong.tsx
+++ b/apps/studio/src/pages/home/releases/EditSong.tsx
@@ -343,7 +343,10 @@ const EditSong: FunctionComponent = () => {
               primaryAction={ () => {
                 deleteSong({
                   redirectToReleases: Boolean(webStudioAlbumPhaseOne),
-                  request: { songId },
+                  request: {
+                    archived: true,
+                    songId,
+                  },
                 });
               } }
               secondaryAction={ () => {
--- a/apps/studio/src/pages/home/releases/SongList.tsx
+++ b/apps/studio/src/pages/home/releases/SongList.tsx
@@ -444,22 +444,15 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
                 <TableCell>
                   <Box sx={ { alignItems: "center", display: "flex" } }>
                     { isSongStale ? (
-                      <Tooltip
-                        title={
-                          "The file couldn't be uploaded. Please try again, " +
-                          `or reach out to ${NEWM_SUPPORT_EMAIL} for further assistance.`
-                        }
-                      >
-                        <CloseIcon
-                          color="error"
-                          sx={ {
-                            height: "24px",
-                            marginLeft: [0, 1],
-                            marginRight: [2, 4],
-                            width: "40px",
-                          } }
-                        />
-                      </Tooltip>
+                      <CloseIcon
+                        color="error"
+                        sx={ {
+                          height: "24px",
+                          marginLeft: [0, 1],
+                          marginRight: [2, 4],
+                          width: "40px",
+                        } }
+                      />
                     ) : (
                       <IconButton
                         sx={ {
@@ -541,23 +534,20 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
                 >
                   { isSongStale ? (
                     <Tooltip
-                      placement="right"
+                      placement="left"
                       title="There was an issue processing your release's audio. Please contact support for assistance."
                     >
-                      <Link
+                      <Button
+                        aria-label="Contact support"
+                        color="music"
                         href={ NEWM_SUPPORT_LINK }
                         rel="noopener noreferrer"
                         target="_blank"
+                        variant="secondary"
+                        onClick={ (event) => event.stopPropagation() }
                       >
-                        <Button
-                          aria-label="Contact support"
-                          color="music"
-                          variant="secondary"
-                          onClick={ (event) => event.stopPropagation() }
-                        >
-                          Support
-                        </Button>
-                      </Link>
+                        Support
+                      </Button>
                     </Tooltip>
                   ) : (
                     <ActionMenuTrigger onClick={ handleMenuOpen(song) } />
--- a/apps/studio/src/pages/home/releases/SongList.tsx
+++ b/apps/studio/src/pages/home/releases/SongList.tsx
@@ -261,6 +261,7 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
 
   const handleMenuClose = () => {
     setMenuAnchorEl(null);
+    setMenuSong(null);
   };
 
   const clearPendingSaleStorage = (songId: string) => {
--- a/apps/studio/src/pages/home/releases/SongList.tsx
+++ b/apps/studio/src/pages/home/releases/SongList.tsx
@@ -398,6 +398,11 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
     setPage(1);
   }, [query]);
 
+  useEffect(() => {
+    setMenuAnchorEl(null);
+    setMenuSong(null);
+  }, [page, query, songData]);
+
   // Keep song in a playing state till the song has been filtered out
   useEffect(() => {
     const isSongFound = !!songData?.find((song) => {
--- a/apps/studio/src/pages/home/releases/SongList.tsx
+++ b/apps/studio/src/pages/home/releases/SongList.tsx
@@ -401,7 +401,11 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
         id: "view-edit",
         label: "View / Edit",
         onClick: () => {
-          navigate(`/home/releases/edit-song/${menuSong.id}`);
+          if (menuSongIsEditable) {
+            navigate(`/home/releases/edit-song/${menuSong.id}`);
+          } else {
+            navigate(`/home/releases/view-details/${menuSong.id}`);
+          }
         },
       },
       {
--- a/apps/studio/src/pages/home/releases/SongList.tsx
+++ b/apps/studio/src/pages/home/releases/SongList.tsx
@@ -53,7 +53,6 @@ import {
   useGetSongsQuery,
 } from "../../../modules/song";
 import {
-  NEWM_SUPPORT_EMAIL,
   NEWM_SUPPORT_LINK,
   PlayerState,
   isSongEditable as isSongEditableUtil,
@@ -76,6 +75,7 @@ const FINAL_STEP_MINTING_PROCESS = [
 ];
 
 export default function SongList({ totalCountOfSongs, query }: SongListProps) {
+  // TODO(webStudioAlbumPhaseOne): Remove flag once flag is retired.
   const { webStudioAlbumPhaseOne } = useFlags();
 
   const navigate = useNavigate();
@@ -293,9 +293,14 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
     if (isErrorMintingStatus) {
       content = (
         <span>
-          An error has occurred. Please reach out to{ " " }
-          <Link href={ `mailto:${NEWM_SUPPORT_EMAIL}` }>
-            { NEWM_SUPPORT_EMAIL }
+          An error has occurred. Please reach out via the{ " " }
+          <Link
+            href={ NEWM_SUPPORT_LINK }
+            rel="noopener noreferrer"
+            target="_blank"
+            onClick={ (event) => event.stopPropagation() }
+          >
+            NEWM Support Portal
           </Link>{ " " }
           for assistance distributing your release.
         </span>
@@ -349,7 +354,7 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
           setIsDeleteModalActive(true);
         },
         tooltip: !isMenuSongDeletable ? <ReleaseDeletionHelp /> : undefined,
-        tooltipPlacement: "right",
+        tooltipPlacement: "left",
       },
     ];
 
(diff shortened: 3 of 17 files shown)


ITurres:
Btw, I kept the "support" option but now into the menu actions for backwards compatibility, this only shows if the selected song was created more than 1200 seconds ago and still has no `streamUrl` (i.e., `upload/processing` likely failed or stalled). This opens the user's email client with a prefilled mailto `support@newm.io`.

Let me know if you'd rather remove it entirely or if you prefer I update the icon/styling.

ebaaa7d3-e019-4c68-a335-a730d93b0913

ITurres:
Also, the View / Edit option currently swaps icons: a pen when editable and an eye when view‑only. This isn't in Figma, but I thought it might be helpful. Let me know if you'd prefer to keep a single icon to match the design.

image

ITurres:
Refined the Action menu’s View/Edit routing to go directly to view-details for non‑editable releases (previous behavior relied on `EditSong` redirect; still worked but caused an extra hop. [8a5d8b0](https://github.com/projectNEWM/newm-web/pull/936/commits/8a5d8b0f0087c8cbf9adbb048b692c916b453b0f)

apps/studio/src/pages/home/releases/SongList.tsx
```diff
@@ -79,6 +99,10 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
   const [page, setPage] = useState(1);
   const [fetchStreamData, fetchStreamDataResp] = useFetchSongStreamThunk();
 
+  const [menuAnchorEl, setMenuAnchorEl] = useState<null | HTMLElement>(null);
+  const [menuSong, setMenuSong] = useState<Song | null>(null);
+  const [isDeleteModalActive, setIsDeleteModalActive] = useState(false);
+
```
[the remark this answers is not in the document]
dmkirshon: @ITurres Might be related, but when resizing the window to update the query/listed releases, when one of the menus is active on a release, can cause the menu component overlay to remain active. It requires an initial click on the page to interact again. Minor edge case.

https://github.com/user-attachments/assets/09ea40ed-af51-4870-8882-b6316a74f6f4

apps/studio/src/pages/home/releases/SongList.tsx
```diff
@@ -231,18 +255,101 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
...
+    await dispatch(
+      songApi.endpoints.deleteSong.initiate({
+        archived: true,
+        songId: menuSong.id,
+      })
+    );
+
+    setIsDeleteModalActive(false);
+    setMenuAnchorEl(null);
+    setMenuSong(null);
```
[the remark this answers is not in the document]
ITurres: We’re intentionally keeping the current behavior. The `deleteSong` mutation already emits a toast on failure ('An error occurred while deleting your song'), so the user is notified even if the modal closes. We chose to close the modal on error because keeping it open places the toast behind the backdrop/overlay, which can make the error less noticeable. If we later want retry-in-place, we can add an inline error message inside the modal - but this later is not in the Figma designs.

apps/studio/src/pages/home/releases/SongList.tsx
```diff
@@ -231,18 +255,101 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
...
+        }
+      } catch (error) {
+        // * Ignores invalid localStorage payloads.
+      }
+    };
+
+    clearSongFromMap(LOCAL_STORAGE_SALE_START_PENDING_KEY);
+    clearSongFromArray(LOCAL_STORAGE_SALE_END_PENDING_KEY);
+    clearSongFromArray(LOCAL_STORAGE_SALE_COMPLETE_PENDING_KEY);
   };
```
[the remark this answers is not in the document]
dmkirshon: @ITurres This localStorage cleanup logic won't apply to any song in the Deletable minting status:

```
MintingStatus.Undistributed,
MintingStatus.StreamTokenAgreementApproved,
MintingStatus.MintingPaymentRequested
```
The sale can only be created after a song has had tokens in a status of:

```
MintingStatus.Minted,
MintingStatus.Released,
```

This seems like code that wouldn't affect any changes. Let me know if this makes sense.

Also, if this code was used, I would have expected this code to be called through `deleteSong` thunk and then `useDeleteSongThunk` would be used like it is in `EditSong.tsx`.

apps/studio/src/pages/home/releases/SongList.tsx
```diff
@@ -269,6 +376,73 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
...
+        ),
+        id: "view-edit",
+        label: "View / Edit",
+        onClick: () => {
+          if (menuSongIsEditable) {
+            navigate(`/home/releases/edit-song/${menuSong.id}`);
+          } else {
+            navigate(`/home/releases/view-details/${menuSong.id}`);
+          }
+        },
```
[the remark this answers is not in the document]
ITurres: Not really a concern here. The icons are just React elements created during render; theyre lightweight, and the menu items are already inside a `useMemo`, so they only re-create when dependencies change (like `menuSongIsEditable`). This isnt a performance issue in practice.

apps/studio/src/pages/home/releases/SongList.tsx
```diff
@@ -231,18 +255,101 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
...
-    event.stopPropagation();
+  const handleMenuOpen =
+    (song: Song) => (event: MouseEvent<HTMLButtonElement>) => {
+      event.stopPropagation();
+      setMenuAnchorEl(event.currentTarget);
+      setMenuSong(song);
+    };
+
+  const handleMenuClose = () => {
+    setMenuAnchorEl(null);
```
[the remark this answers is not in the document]
ITurres: Addressed 🚀

packages/elements/src/lib/styled/GradientDashedOutline.tsx
```diff
@@ -26,7 +26,9 @@ const GradientDashedOutline = styled(Box, {
   backgroundImage: `linear-gradient(${theme.colors.grey700}, ${theme.colors.grey700}), ${theme.gradients[gradient]}`,
   backgroundOrigin: "padding-box, border-box",
   border: `2px dashed ${theme.colors.grey700}`,
+  borderRadius: 5,
 
-  borderRadius: 4,
+  borderWidth: "2.5px",
```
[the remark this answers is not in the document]
ITurres: Addressed 🚀

apps/studio/src/pages/home/releases/SongList.tsx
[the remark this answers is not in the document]
ITurres: Thank you @dmkirshon ! Addressed 🚀

apps/studio/src/pages/home/releases/SongList.tsx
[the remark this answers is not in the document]
ITurres: @dmkirshon Thank you for pointing this out 🙏🏼. you're right that with the current rules, deletable statuses (Undistributed on the frontend) won't overlap with sale pending statuses (that seems to occur after minting/distribution), so the localStorage cleanup is likely unnecessary in a normal flow. I added it defensively in case a song ID is in those pending‑sale keys (e.g., earlier states or edge cases), but it's not strictly necessary given the current minting statuses. So I can remove the cleanup. thoughts?

On the thunk: I intentionally didn't use `useDeleteSongThunk` in the releases table because it forces a redirect to `/home/library`. There I intended to keep the user on the releases list and rely on RTK invalidation instead. But it is true that in the end we'll likely need to use the thunk , so I could use `isProd` to env‑gate the routing in the thunk to take the user to `/home/releases` for now, since we can't really use an LD flag there. and update `songList` to use it.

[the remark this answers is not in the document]
dmkirshon: Agree, remove the `localStorage` clean up as there is no edge case currently for the Sale logic to be accessible while a user can delete their song/release. When marketplace tab is accessible the Delete icon is already disabled the user and any song deletion would have to go through support/backend.

If you are able to send the LD flag, or redirectPath, via a parameter or another clean way, I would integrate that now for the `deleteSong` thunk. We will need to figure out a clean way to do it across a few other thunks, like `uploadSong`, that also push to Library currently. Those will be addressed in the Navigation update ticket.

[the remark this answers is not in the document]
ITurres: Great, thanks @dmkirshon! moving forward with the `localStorage` clean up removal. **[DONE]** ✅ 

About the LD + thunk: yes, we can pass the flag as an optional parameter to the thunk whille keeping it backward‑compatible for existing callers. For the current `deleteSong` I added a small wrapper type so current usage still works, and the new redirect behavior is only applied when the flag is provided. **[DONE]** ✅

apps/studio/src/pages/home/releases/SongList.tsx
dmkirshon: The tooltip here should be removed as it has mixed messages with the support button tooltip below.

apps/studio/src/pages/home/releases/SongList.tsx
```diff
@@ -448,30 +540,27 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
...
+                      >
+                        <Button
+                          aria-label="Contact support"
+                          color="music"
+                          variant="secondary"
+                          onClick={ (event) => event.stopPropagation() }
+                        >
+                          Support
+                        </Button>
+                      </Link>
```
dmkirshon: The custom NEWM Button leverages the MUI button, if href is provided it acts as an `a` element at it's base. The `Link` attributes can be moved to the `Button` and it should function the same way as the `Link` wrapper.

apps/studio/src/pages/home/releases/SongList.tsx
```diff
@@ -448,30 +540,27 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
                   } }
                 >
                   { isSongStale ? (
-                    <Button
-                      color="music"
-                      variant="secondary"
-                      onClick={ (event) => handleEmailSupport(event, song.id) }
+                    <Tooltip
+                      placement="right"
```
dmkirshon: Is there a reason for placement `"right"` when it's at the end of the table? Shouldn't it be placed to the `"left"` based on location?

apps/studio/src/pages/home/releases/SongList.tsx
```diff
@@ -231,16 +251,37 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
...
-      encodeURIComponent(`The following Song ID failed to encode: ${songId}`);
+    if (pendingDeleteSong.id === playerState.currentPlayingSongId) {
+      stopSong();
+    }
 
-    window.location.href = mailtoLink;
+    await deleteSong({
+      redirectToReleases: Boolean(webStudioAlbumPhaseOne),
+      request: {
+        archived: true,
```
dmkirshon, line 276: `Archived: true` is the default in the endpoint, is this just for readability?

https://github.com/projectNEWM/newm-web/blob/08495ce888be87b78a187e5dee58891003235996/apps/studio/src/modules/song/api.ts#L203

apps/studio/src/pages/home/releases/EditSong.tsx
```diff
@@ -332,7 +341,13 @@ const EditSong: FunctionComponent = () => {
           { isDeleteModalActive && (
             <DeleteSongModal
               primaryAction={ () => {
-                deleteSong({ songId });
+                deleteSong({
+                  redirectToReleases: Boolean(webStudioAlbumPhaseOne),
+                  request: {
+                    archived: true,
```
dmkirshon, line 348: Same as SongList.tsx comment about `archived: true`.

https://github.com/projectNEWM/newm-web/pull/936/changes#r2712156579

apps/studio/src/pages/home/releases/SongList.tsx
ITurres: Yes, agreed, thanks david! 🚀

apps/studio/src/pages/home/releases/SongList.tsx
ITurres: Indeed, it does function the same! removing the `Link` now 💪🏼 **[DONE ✅]**

apps/studio/src/modules/song/thunks.ts
```diff
@@ -562,17 +562,26 @@ export const getUserWalletSongs = createAsyncThunk(
  * Request to delete user song. If successful, navigate to
  * library and fetch new songs.
  */
```
dmkirshon, line 564: This moves the Doc description off of the `deleteSong` function. I think that was its other purpose, not only for delineation from the other thunks.

I think a comment/note should also be made that this type/conditional flow should be cleaned up when the LD flag `webstudioalbumphaseone` is removed. However, if we are tracking these changes another way, through the Jira ticket, then it isn't fully necessary. Just don't want to lose track of all these required migrations when we get to them.

apps/studio/src/pages/home/releases/SongList.tsx
ITurres: Good catch 🚀! No reason, I missed it. Switching to `left` now. **[DONE ✅]**

apps/studio/src/pages/home/releases/SongList.tsx
ITurres: Good call @dmkirshon, the API defaults archived to true, so it's indeed not required. I set it explicitly in both `SongList` and then in `EditSong` for clarity, since the thunk is named `delete` but the action is actually a PATCH/soft‑delete. But no problem on removing it if you prefer the defaulted behavior only.

apps/studio/src/modules/song/thunks.ts
```diff
@@ -562,17 +562,26 @@ export const getUserWalletSongs = createAsyncThunk(
...
+type DeleteSongRedirectVariant = {
+  redirectToReleases?: boolean;
+  request: DeleteSongRequest;
+};
```
dmkirshon, line 569: Just wondering why the original `DeleteSongRequest` type wasn't modified with the optional `redirectToReleases`?

apps/studio/src/modules/song/thunks.ts
ITurres: Agree @dmkirshon , I think we can take two approaches on how to track these changes, we can add a brief `TODO:` comment near the places where flags were used, in this case noting the redirect variant should be removed when `webStudioAlbumPhaseOne` is gone, and also track it in Jira as you suggested, to avoid losing file/line context. But if you'd prefer not todo's comments on files then we can list (with links also) the modified files in a jira ticket instead.

apps/studio/src/modules/song/thunks.ts
ITurres: @dmkirshon I kept `DeleteSongRequest` unchanged because `redirectToReleases` isn't part of the actual API payload, it's just for UI control-flow in the thunk. I didnt want to pollute the request type or imply the server consumes it. Once the LD path goes away we can drop the wrapper and revert to the original type.

apps/studio/src/pages/home/releases/SongList.tsx
ITurres: @dmkirshon quick question on the tooltip at [getTooltipContent](https://github.com/projectNEWM/newm-web/blob/master/apps/studio/src/pages/home/releases/SongList.tsx#L412): it currently shows the old support email for minting errors, but it also shows a useful (?) declined‑distribution message. Maybe I could: update the error copy to the new Support portal link and keep the declined message as is. Thoughts?

apps/studio/src/modules/song/thunks.ts
dmkirshon: TODO comment sounds good, no problem with them. If we refer the `webStudioAlbumPhaseOne` even better, so it will be easy to find/review any migration changes. Whatever we do, as long is it's consistent for the future migration/clean up.

apps/studio/src/pages/home/releases/SongList.tsx
dmkirshon: We can leave it for clarity since it isn't obvious without digging down two-three files.

apps/studio/src/modules/song/thunks.ts
ITurres: Ok great @dmkirshon! we can prefix the comment like `TODO(webStudioAlbumPhaseOne): ...` so then it's easy to locate them with grep.

apps/studio/src/pages/home/releases/SongList.tsx
```diff
@@ -269,10 +310,61 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
...
+        disabled: !isMenuSongDeletable,
+        icon: <DeleteIcon fontSize="small" />,
+        id: "delete",
+        label: "Delete",
+        onClick: () => {
+          setPendingDeleteSong(menuSong);
+          setIsDeleteModalActive(true);
+        },
+        tooltip: !isMenuSongDeletable ? <ReleaseDeletionHelp /> : undefined,
+        tooltipPlacement: "right",
```
dmkirshon: Not sure if this was caught as well, this is listed as `"right"`, not sure if there are others. This will be at the end of the table on the right, might be better to keep defaulted as `"left"`.

apps/studio/src/pages/home/releases/SongList.tsx
dmkirshon: Yes I think that would be a small win as well, mention it on the Jira ticket and it will be good.

apps/studio/src/pages/home/releases/SongList.tsx
```diff
@@ -252,9 +293,14 @@ export default function SongList({ totalCountOfSongs, query }: SongListProps) {
...
-          An error has occurred. Please reach out to{ " " }
-          <Link href={ `mailto:${NEWM_SUPPORT_EMAIL}` }>
-            { NEWM_SUPPORT_EMAIL }
+          An error has occurred. Please reach out via the{ " " }
+          <Link
+            href={ NEWM_SUPPORT_LINK }
+            rel="noopener noreferrer"
+            target="_blank"
+            onClick={ (event) => event.stopPropagation() }
+          >
```
ITurres, line 302: Note: using MUI `Link` here because our `@newm-web/elements` `Link` doesn't currently accept `target`, and we need `target="_blank"` for the Support Portal link.

dmkirshon: Is the `stopPropagation` needed here for a `Link` vs for a `Button`?

dmkirshon: If it's within the tooltip it shouldn't interact with any other Page navigation/logic?

dmkirshon: Nevermind, it does cause propagation! Good catch!

dmkirshon (approved):
All comments and discussions resolved. PR is good to merge.

PR #937, opened 2026-01-21
[STUD-538][Enhancement]: Transition from "Library" to "Releases" and consolidate navigation links

ITurres:
# Pull Request — Issue #[STUD-538](https://projectnewm.atlassian.net/browse/STUD-538)

## Overview

> Gate Releases routing behind `webStudioAlbumPhaseOne`, redirect Library routes
> to Releases when enabled, and align navigation copy/flows.
> Centralize `SongRouteParams` in common types, relocate shared playback
> components out of `components/library`.
> Route login to `/home` so Home decides the flag-based landing.

---

## Files Summary

> **Total Changes:** 32 files

<details>
<summary>Added (0)</summary>

</details>

<details>
<summary>Modified (28)</summary>

- **`newm-web/apps/studio/src/pages/home/Home.tsx`**
  - — Redirect Library routes to Releases when the flag is on and default home route to Releases.
- **`newm-web/apps/studio/src/pages/home/SideBar.tsx`**
  - — Toggle label/link between Library and Releases behind the flag.
- **`newm-web/apps/studio/src/components/paypal/PayPalSuccessSession.tsx`**
  - — Update copy to reflect `Releases` vs `Library` under the flag.
- **`newm-web/apps/studio/src/modules/session/thunks.ts`**
  - — Navigate to `/home` after login instead of `/home/upload-song`. So `Home.tsx` decides the flag-based landing.
- **`newm-web/apps/studio/src/modules/song/thunks.ts`**
  - — Simplify delete flow and remove releases redirect variant.
- **`newm-web/apps/studio/src/common/types.ts`**
  - — Add shared `SongRouteParams`.
- **`newm-web/apps/studio/src/components/index.ts`**
  - — Re-export playback components from `components/releases`.
- **`newm-web/apps/studio/src/pages/home/library/*.tsx`**
  - — Update `SongRouteParams` imports to shared common types.
- **`newm-web/apps/studio/src/pages/home/releases/*.tsx`**
  - — Update `SongRouteParams` imports and delete flow usage.
- **`newm-web/apps/studio/src/pages/home/uploadSong/*.tsx`**
  - — Update `SongRouteParams` imports to shared common types.

</details>

<details>
<summary>Deleted (2)</summary>

- **`newm-web/apps/studio/src/pages/home/library/types.ts`**
  - — Removed redundant `SongRouteParams` definition.
- **`newm-web/apps/studio/src/pages/home/releases/types.ts`**
  - — Removed duplicate `SongRouteParams` definition.

</details>

<details>
<summary>Renamed (2)</summary>

- **`newm-web/apps/studio/src/components/library/IconStatus.tsx`**
  - — Moved to `apps/studio/src/components/releases/IconStatus.tsx`.
- **`newm-web/apps/studio/src/components/library/SongStreamPlaybackIcon.tsx`**
  - — Moved to `apps/studio/src/components/releases/SongStreamPlaybackIcon.tsx`.

</details>

---

## Impact

- When `webStudioAlbumPhaseOne` is on, `/home` and `/home/library/*` route to Releases.
- Login flows now land on `/home` instead of `/home/upload-song`.
- `SongRouteParams` is centralized and used consistently across Library/Releases/Upload.
- Shared playback components live under `components/releases` (no more `components/library`).
- No changes to server API contracts or dependencies.

---

## Testing

- run: `npx nx test studio`
  - All tests pass.

---

## Related Issues

- Closes #[STUD-538](https://projectnewm.atlassian.net/browse/STUD-538)

---

## Dependencies

- No new dependencies.

---

## Demo  >>UPDATED<<

https://github.com/user-attachments/assets/8d08d0f0-b428-4b9e-a07a-f0a3a1cc034d

---

## Additional Notes

- N/A.

--END--

[STUD-538]: https://projectnewm.atlassian.net/browse/STUD-538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[STUD-538]: https://projectnewm.atlassian.net/browse/STUD-538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/elements/src/lib/SideBarNavLink.tsx
```diff
@@ -1,7 +1,8 @@
 import { CSSProperties, FunctionComponent, ReactNode } from "react";
-import { Link, useMatch, useResolvedPath } from "react-router-dom";
+import { useMatch, useResolvedPath } from "react-router-dom";
 import { Box, Stack, SvgIconProps } from "@mui/material";
 import theme from "@newm-web/theme";
+import Link from "./Link";
```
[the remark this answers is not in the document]
dmkirshon: @ITurres Looking over this again, this is a custom element version of `Link`, you can revert back to  `"react-router-dom"`. I got wrapped around the use of `Link` imports earlier, but this is one place where it makes sense to use the routing `Link`.

Thank you for working through it.

apps/studio/src/pages/home/Home.tsx
```diff
@@ -105,7 +115,20 @@ const Home: FunctionComponent = () => {
           <Route element={ <Navigate to="upload-song" replace /> } path="" />
```
dmkirshon: There are a few places where we default navigation to `upload-song` instead of `Releases`. Is this still intended behavior for phase 1? 
- Home -> upload-song
- Login -> upload-song

apps/studio/src/pages/home/Home.tsx
dmkirshon, line 1: There is a Library type, `import { SongRouteParams } from "../library/types";` that should probably refactored to `apps/studio/src/common/types.ts`. Thoughts? 

`apps/studio/src/pages/home/uploadSong/AdvancedSongDetails.tsx`

`apps/studio/src/pages/home/uploadSong/UploadSong.tsx`

`apps/studio/src/pages/home/uploadSong/BasicSongDetails.tsx`

`apps/studio/src/pages/home/uploadSong/SongDetails.tsx`

`apps/studio/src/pages/home/uploadSong/ReviewAndMint.tsx`

The other reference to library I found was `apps/studio/src/components/library` that has shared components for the `SongStreamPlaybackIcon` and `iconStatus`. We can clean this one up in Phase 2, but just wanted to note it.

dmkirshon (changes requested):
We should revert the `SideBarNavLink` fix, also provided a few follow up comments.

packages/elements/src/lib/SideBarNavLink.tsx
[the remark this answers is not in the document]
ITurres: reverting now... 💯

apps/studio/src/pages/home/Home.tsx
ITurres: @dmkirshon right, under the current phase 1 flag we could navigate the user to `releases` to satisfy `Existing routes previously pointing to “Upload a Song” should be verified to navigate correctly under the Phase 1 LD Flag when enabled (true).` great catch 🚀

apps/studio/src/pages/home/Home.tsx
ITurres: @dmkirshon I confirmed theres a library-only `SongRouteParams` type yes (and a separate releases copy too), so centralising it in `apps/studio/src/common/types.ts` is the right direction. This could be considered out of scope for the current ticket, but since Phase 2 is where we're already cleaning up library-related items and it's easy to miss these scattered types, I could go ahead and addressed it in this branch so we dont have to track it later.

Also saw the `apps/studio/src/components/library` shared components note. If we address the type imports, then I can also move/rename that folder now (there are no many references) and update imports in this PR to reduce Phase 2 cleanup as well.

dmkirshon: I would say yes on both, will set us up for clean up in the next phase.

packages/elements/src/lib/SideBarNavLink.tsx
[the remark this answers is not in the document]
ITurres: **[DONE] ✅**

apps/studio/src/pages/home/Home.tsx
ITurres: **[DONE] ✅**

apps/studio/src/pages/home/Home.tsx
ITurres: both **[DONE] ✅**

dmkirshon, line 87: I think we are going to need to update the code similar to this logic with:

Since we know the flag is present in all environments and we aren't diverging to having unique flag environments:

`if (webStudioAlbumPhaseOne === undefined) return null;`

or more robust and safe:

 ```const ldClient = useLDClient();
  const [areFlagsReady, setAreFlagsReady] = useState(false);

  useEffect(() => {
    if (!ldClient) return;
    let mounted = true;

    ldClient.waitForInitialization().then(() => {
      if (mounted) setAreFlagsReady(true);
    });

    const onReady = () => mounted && setAreFlagsReady(true);
    ldClient.on("ready", onReady);

    return () => {
      mounted = false;
      ldClient.off("ready", onReady);
    };
  }, [ldClient]);

  if (!areFlagsReady) return null;
```

When navigating to `/home` it still loads `/upload-song` for me, the LD flag doesn't load in time for the conditional navigational route replace. I also see the "404" you mentioned and notice that the Sidebar navigation has a staggered load of "Upload A Song" and "Library", only after the LD flag loads does everything switch over to `/releases`. Either choice above should allow time for LD to load so that the Routes/Navigation can load without stutters.

File: apps/studio/src/pages/home/Home.tsx [hunk 3]
```diff
@@ -102,10 +112,31 @@ const Home: FunctionComponent = () => {
...
+                  replace
+                />
+              ) : (
+                <Library />
+              )
+            }
+            path="library/*"
+          />
 
           { isReleasesPageEnabled && (
```
dmkirshon: Is this variable needed? We should be consistent and either label all the webStudioAlbumPhaseOne using this variable or remove it altogether.

apps/studio/src/pages/home/Home.tsx
dmkirshon: I think I was doing some extra manipulation in the local environment that was causing this issue to be exaggerated. The proposed fix might cause other `null` render issues depending on network/LDclient loads. I would be ok to push forward since it does seem more like an initial cache issue, likely a minor edge case. I will leave it up to you.

dmkirshon (approved):
Looks good as is, the minor cache issue shouldn't block. The variable clean up is also minor but would be good to address for consistency.

apps/studio/src/pages/home/Home.tsx
ITurres: yes I do see the concern as well, we could guard the LD for readiness which could smooth the first load sort of flickering, but also as you mentioned we could have a null render issue if the LD is slow or it's block.. I think that since we are lifting the flag soon after verification, we could keep it as is and avoid introducing that null risk + more code. But it is something to consider for phase 2 in case we see it happening over and over.

So the more robust and safe code you shared, we could consider moving it to a hook, and we reuse it everywhere when needed 🚀

apps/studio/src/pages/home/Home.tsx
ITurres: Good catch! 😎 yes I'll remove it and rely on `webStudioAlbumPhaseOne` flag only.

ITurres: **DONE ✅**

